PR #14, opened 2023-06-06
Fix GCC build error by correctly including `<utility>` header

xkevio:
This is needed to not rely on implementation defined behavior regarding transitive header includes, here for std::integer_sequence.


wheremyfoodat:
Could you add type_traits for std::integral_constant too

wheremyfoodat:
Lookin good

PR #15, opened 2023-06-06
Vertex shaders & vertex loader -> x64 JIT

wheremyfoodat:


wheremyfoodat:
image
OK maybe we're going in the wrong direction

georgemoralis:
Hmm mostly dummy code atm but I am kinda excited!

wheremyfoodat:
🥰

Wunkolo:
Excited for this one! Lots of potential speed with stuff like AVX2/AVX512.

src/core/PICA/dynapica/shader_rec.cpp
```diff
@@ -0,0 +1,31 @@
...
+void ShaderJIT::reset() {
+	cache.clear();
+}
+
+void ShaderJIT::prepare(PICAShader& shaderUnit) {
+	shaderUnit.pc = shaderUnit.entrypoint;
+	// We construct a shader hash from both the code and operand descriptor hashes
+	// This is so that if only one of them changes, we still properly recompile the shader
+	// This code is inspired from how Citra solves this problem
+	Hash hash = shaderUnit.getCodeHash() ^ shaderUnit.getOpdescHash();
```
Wunkolo, line 14: This is a total nit-pick but xor is a commutative operation so behavior like `a^b == b^a` could produce the same hash here causing a possible collision-hazard. It's ridiculously low of course but you might want to do something like `std::rotl(x, 1) ^ y` to be more correct here.

wheremyfoodat: Very good observation. I'll think about it for a sec and change it.

wheremyfoodat: Added

wheremyfoodat:
Superseded in #39

PR #16, opened 2023-06-10
Fix cmake out-of-source build instructions

Wunkolo:
These directions are incorrectly using `cmake .` when it should be
`cmake ..`.


wheremyfoodat (approved):

wheremyfoodat:
Thanks!

PR #18, opened 2023-06-10
Add `Helpers:{getBits,getBit}`

Wunkolo:
`constexpr` functions for extractint bitfields that lends itself a bit
better to emitting instructions like `bextr` on x86 or `ubfx` on arm64.
And may subjectively make things a bit more readable.
"Extract `5` bits" rather than `& 0x1F`.

### DIFF
--- a/include/PICA/shader.hpp
+++ b/include/PICA/shader.hpp
@@ -128,14 +128,14 @@ class PICAShader {
 		bool negate;
 
 		if constexpr (sourceIndex == 1) { // SRC1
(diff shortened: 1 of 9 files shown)


wheremyfoodat:
Looks good, will look at the affected areas to make sure everything is good and then merge

wheremyfoodat (approved):
Looked at it a bit with my eyes, seems good, will try OoT to see it plays properly to avoid any oversights and merge

PR #19, opened 2023-06-10
Add `FourCC`/`_u32` helper function/literals

Wunkolo:
Portable code for turning string-based four-character-codes like `"BLAH"_u32` into a 32-bit integer at compile-time.

### DIFF
--- a/include/helpers.hpp
+++ b/include/helpers.hpp
@@ -141,6 +141,28 @@ constexpr size_t operator""_GB(unsigned long long int x) {
     return 1024_MB * x;
 }
 
+// Utility user-literal/metaprogramming-type for turning four-character-codes
+// into 32-bit integers at compile-time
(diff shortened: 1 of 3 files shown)


Wunkolo:
Testbed of this working can be seen here:
https://godbolt.org/z/55c4v43TY

wheremyfoodat:
Hmmm, so, I've kind of got mixed feelings about this because the FourCC struct itself is good, though I kind of want to avoid the punning in the NCSD/NCCH/etc headers, mainly cause the codebase has a bit way too much punning already

wheremyfoodat:
Also as CI thinks, std::endian might be a biiit too niche of a C++20 feature to use atm.

Wunkolo:
For sure, yeah I think those are the only two places that even use FourCC. Prob not worth it 👌

PR #20, opened 2023-06-11
Implement `SaveDataArchive::createFile`

Wunkolo:
Based on `ExtSaveDataArchive::createFile`. Gets some games farther that
try to create save-data upon boot.

### DIFF
--- a/src/core/fs/archive_save_data.cpp
+++ b/src/core/fs/archive_save_data.cpp
@@ -5,9 +5,6 @@
 namespace fs = std::filesystem;
 
 FSResult SaveDataArchive::createFile(const FSPath& path, u64 size) {
-	if (size == 0)
-		Helpers::panic("SaveData file does not support size == 0");
-
 	if (path.type == PathType::UTF16) {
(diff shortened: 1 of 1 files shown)


wheremyfoodat (dismissed):
Very nice, any game I should try with this?

Wunkolo:
Currently only tested it with `Picross 3D - Round 2` which creates a `SAVEDATA` file upon first-boot.
Needs this line to be commented out as well:
https://github.com/wheremyfoodat/Panda3DS/blob/fd600a59c164e3730a54e6368642f7da3de1bcd2/src/core/PICA/gpu.cpp#L262

It gets farther in its initialization before halting at [APT:PreloadLibraryApplet](https://www.3dbrew.org/wiki/APT:PreloadLibraryApplet)

wheremyfoodat:
This needs some changes (eg size == 0 is supported on SaveData and some error codes might be wrong). Will try fixing and merging

wheremyfoodat (approved):

PR #22, opened 2023-06-11
Implement more DSP pipe stuff, stub APT::PreloadLibraryApplet for Picross 3D

wheremyfoodat:


Wunkolo:
Can confirm this gets me in!
image

wheremyfoodat:
@Wunkolo Wouldn't it be amazing if we had touch screen support when you click on the bottom screen part, but didn't actually have the bottom screen?

PR #29, opened 2023-06-15
Add `.clang-format`

Wunkolo:
Used the command `clang-format -i tests/**/*.c src/**/*.cpp include/**/*.hpp` to process all of the project's files.
Using the one from [pcsx-redux](https://github.com/grumpycoders/pcsx-redux/blob/main/src/.clang-format)as a basis and then made additional edits to be the _least_ disruptiveto the pre-existing formatting patterns.

Open to additional edits.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -5,10 +5,10 @@ AccessModifierOffset: -2
 TabWidth: 4
 NamespaceIndentation: All
 UseTab: ForContinuationAndIndentation
-AllowShortEnumsOnASingleLine: false
+AllowShortEnumsOnASingleLine: true
 AllowShortCaseLabelsOnASingleLine: true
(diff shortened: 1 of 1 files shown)


wheremyfoodat:
Could you try PRing this without the clang-format -i ** part first? I think it would be easier to ensure clang-format does not break something terribly by doing that and formatting files little by little now and then.

wheremyfoodat (approved):

wheremyfoodat:
Switching the "allow short x in single line" options to true because a lot of code is like this https://github.com/wheremyfoodat/Panda3DS/blob/2e4ae1328ae0c7d336688157eee2891eb0ba2345/include/services/service_manager.hpp#L75-L95

PR #33, opened 2023-06-16
Use `std::span` to pass vertex data

Wunkolo:
Starts utilizing [std::span](https://en.cppreference.com/w/cpp/container/span) to indicate a non-owning view of a contiguous array of elements rather than `T* data, usize count`.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,6 +1,6 @@
 BasedOnStyle: Google
 IndentWidth: 4
-ColumnLimit: 120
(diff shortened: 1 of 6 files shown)


wheremyfoodat:
Can you merge current master into this rq

wheremyfoodat (changes requested):
boooomb

wheremyfoodat (approved):
This codebase was greek at first, now it's becoming span-ish
🥁

PR #36, opened 2023-06-16
MMU-accelerated fastmem (Using Duckstation's memory arena impl)

wheremyfoodat:


wheremyfoodat:
> @wheremyfoodat Memory mappings are forced to have a read permission right now, please make it configurable

OK

wheremyfoodat:
After speaking with the author of DS it's probably best to switch to a memory arena like PCSX2's which can properly do things like 4KB granularity pages on Windows

PR #37, opened 2023-06-16
hle: Add proper type for result code

marysaka:
This should clean up all HLE errorcode in the codebase.

I didn't removed Rust::Result as this should be a cleanup for another iteration.

### DIFF
--- a/include/result/result_common.hpp
+++ b/include/result/result_common.hpp
@@ -1,0 +1,206 @@
+#pragma once
+
+#include <climits>
+#include <cstdint>
+#include <type_traits>
+
+namespace Result {
+	enum class HorizonResultLevel : uint32_t {
+		Success = 0,
+		Info = 1,
+		Status = 25,
+		Temporary = 26,
+		Permanent = 27,
+		Usage = 28,
+		Reinitialize = 29,
+		Reset = 30,
+		Fatal = 31,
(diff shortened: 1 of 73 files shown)


wheremyfoodat (approved):
Good to go, I just need to double check the errors once more

include/result/result_common.hpp
```diff
@@ -0,0 +1,206 @@
...
+		static const uint32_t ModuleOffset = DescriptionOffset + DescriptionBits;
+		static const uint32_t SummaryOffset = ModuleOffset + ModuleBits + ReservedBits;
+		static const uint32_t LevelOffset = SummaryOffset + SummaryBits;
+
+		static_assert(DescriptionBits + ModuleBits + SummaryBits + LevelBits + ReservedBits == sizeof(uint32_t) * CHAR_BIT, "Invalid Result size");
+
+		uint32_t m_value;
+
+		constexpr inline uint32_t getBitsValue(int offset, int amount) {
+			return (m_value >> offset) & ~(~0 << amount);
```
wheremyfoodat, line 152: This could use the `getBits<offset, amount>` function @Wunkolo added, though it's understandable if you'd rather not in order to make this part of the codebase "portable" (ie easy to use outside the codebase as well)

PR #38, opened 2023-06-16
misc: Improve kernel syscall checks

marysaka:
Fix some stuffs around.

Based on kernel 11.14

Depends on #37

### DIFF
--- a/include/kernel/kernel.hpp
+++ b/include/kernel/kernel.hpp
@@ -104,6 +104,8 @@ private:
 	std::string getProcessName(u32 pid);
 	const char* resetTypeToString(u32 type);
 
+	int copyStringFromUser(u8 *dst, u32 src, u32 size);
+
 	MAKE_LOG_FUNCTION(log, kernelLogger)
 	MAKE_LOG_FUNCTION(logSVC, svcLogger)
 	MAKE_LOG_FUNCTION(logThread, threadLogger)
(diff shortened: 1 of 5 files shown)


include/kernel/kernel.hpp
```diff
@@ -104,6 +104,8 @@ class Kernel {
 	std::string getProcessName(u32 pid);
 	const char* resetTypeToString(u32 type);
 
+	int copyStringFromUser(u8 *dst, u32 src, u32 size);
```
Wunkolo, line 107: Rather than `int` maybe it's time for `i32`/`i16`/`i8`-types.

wheremyfoodat: Yeah s32 sounds good

src/core/kernel/kernel.cpp
```diff
@@ -162,12 +181,23 @@ void Kernel::getSystemTick() {
...
 
 // Result OutputDebugString(const char* str, s32 size)
-// TODO: Does this actually write an error code in r0 and is the above signature correct?
 void Kernel::outputDebugString() {
 	const u32 pointer = regs[0];
-	const u32 size = regs[1];
+	const s32 size = regs[1];
+
+	if (size < 0 || ~pointer < size) {
+		regs[0] = Result::OS::OutOfRange;
```
wheremyfoodat, line 189: What does the `~pointer < size` do?

marysaka:
I'm not happy about the current PR, I will try to rethink this and open another PR later.

PR #43, opened 2023-06-17
Add PICA texel-format and topology types

Wunkolo:
Slowly stepping the codebase towards having renderer-agnostic types and keeping the translation of PICA-types to OpenGL/VK/DX/Software/etc to the renderer-backend.

### DIFF
--- a/include/PICA/regs.hpp
+++ b/include/PICA/regs.hpp
@@ -128,3 +128,50 @@ namespace PICAInternalRegs {
 		VertexShaderOpDescriptorData7 = 0x2DD,
 	};
 }
+
+enum class PICAColorFmt : u32 {
+	RGBA8 = 0,
+	BGR8 = 1,
+	RGB5A1 = 2,
+	RGB565 = 3,
+	RGBA4 = 4,
+
+	// Technically selectable, but their function is unknown
+	Unknown5 = 5,
+	Unknown6 = 6,
+	Unknown7 = 7,
+};
+
+enum class PICADepthFmt : u32 {
+	Depth16 = 0,
+	Unknown1 = 1,  // Technically selectable, but function is unknown
+	Depth24 = 2,
+	Depth24Stencil8 = 3,
+};
+
+// Size occupied by each pixel in bytes
+
+// All formats are 16BPP except for RGBA8 (32BPP) and BGR8 (24BPP)
+inline constexpr usize sizePerPixel(PICAColorFmt format) {
+	switch (format) {
+		case PICAColorFmt::BGR8: return 3;
+		case PICAColorFmt::RGBA8: return 4;
+		default: return 2;
+	}
+}
+
+inline constexpr usize sizePerPixel(PICADepthFmt format) {
+	switch (format) {
+		case PICADepthFmt::Depth16: return 2;
+		case PICADepthFmt::Depth24: return 3;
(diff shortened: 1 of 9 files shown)


wheremyfoodat:
Maybe switch "Format" to "Fmt" in the ColourFormat/DepthFormat names and PrimitiveTopology to Prim(itive)Type for brevity? Otherwise yeah the GL renderer should definitely not "own" these definitions.

wheremyfoodat:
(I didn't actually know what topology means before this)

include/PICA/regs.hpp
```diff
@@ -127,4 +127,51 @@ namespace PICAInternalRegs {
 		VertexShaderOpDescriptorData6 = 0x2DC,
 		VertexShaderOpDescriptorData7 = 0x2DD,
 	};
-}
\ No newline at end of file
+}
```
wheremyfoodat, line 130: Maybe it's better to add a `namespace PICA {...}` for this stuff so that sizePerPixel is not a global function and that we can drop the "PICA" part of PICAColorFmt/PICADepthFmt (which is reminiscent of C namespacing)

wheremyfoodat: If you agree and don't feel like doing it atm I can do it for you

Wunkolo: I agree, wasn't sure how much "C-programming in C++" we wanted.

wheremyfoodat: okie okie

wheremyfoodat:
This merges the ColourBuffer::Formats enum and the Texture format enum which makes sense because the enum values themselves seem to match for the documented framebuffer formats, however I am wary of doing that because having the colour buffer code see formats such as "ETC1A4" (which is exclusive a texture format and not a framebuffer format) might be dangerous.

Maybe split them into separate `TextureFmt` and `ColorFmt` for now?

Wunkolo:
Sure, though this would double-up any `formatTostring`, `sizePerPixel`, conversions, and other utility-code for two separate enum-definitions, would having `ColourBuffer` produce an error on unsupported formats be just as acceptable? Having "you're doing something weird"-code during exceptional moments, can be a bit better than maintaining two enums with overlapping values in this case. [GPUREG_COLORBUFFER_FORMAT](https://www.3dbrew.org/wiki/GPU/Internal_Registers#GPUREG_COLORBUFFER_FORMAT) also makes any invalid incoming formats unrepresentable anyways.

wheremyfoodat:
It seems like it makes them unrepresentable but it turns out that DMP actually hates you and the format option is 3 bits wide so there's 8 possible encodings.

I think the duplication might be the most scalable and least error prone way of doing it in case we discover any weird behavior for the undocumented ones or something

Wunkolo:
For sure, can always make the utility functions for the framebuffer-format ones alias to the texture-format ones to re-use code too.

wheremyfoodat (approved):
@Wunkolo Looks good to me I believe. Approving and putting off merging for a small while in case @fleroviux or @skylersaleh believe merging this would make their ongoing GPU bonking harder due to merge conflicts.

skylersaleh:
If @fleroviux is good with merging I'm good 👍

fleroviux:
I think any merge conflicts should be manageable. Looks good to me!

PR #44, opened 2023-06-18
Work around compilation errors on macOS

skylersaleh:
Allows compilation on native macOS build system.
Add workaround helper for bit_cast and missing #include <bit> header. 

This PR doesn't have the detection and replacement of the bit_cast with std::bit_cast, which is left as a TODO item.


wheremyfoodat:
It is really beautiful seeing Linux CI take 7 mins, Windows CI taken 10 mins, and MacOS taking 23 mins.

PR #45, opened 2023-06-19
feat: Add support for encrypted NCSD

marysaka:
This adds support for encrypted NCSD files.

We loads keys from `<data_dir>/Alber/sysdata/aes_keys.txt`.

NOTE: We do differ from other emulators by not hard-coding the `generator` key, it's the user responsibility to provide it in `aes_keys.txt`.

NOTE: This is missing NCCH version 1 support and seeded crypto but that would be enough for now.

### DIFF
--- a/include/crypto/aes_engine.hpp
+++ b/include/crypto/aes_engine.hpp
@@ -1,0 +1,167 @@
+#pragma once
+
+#include <array>
+#include <cstring>
+#include <cstdint>
+#include <climits>
+#include <filesystem>
+#include <optional>
+
+#include "helpers.hpp"
+
+namespace Crypto {
+	constexpr std::size_t AesKeySize = 0x10;
+
+	using AESKey = std::array<uint8_t, AesKeySize>;
+
+	template<std::size_t N>
+	static std::array<uint8_t, N> rolArray(const std::array<uint8_t, N>& value, std::size_t bits) {
+		const auto bitWidth = N * UINT8_WIDTH;
+
+		bits %= bitWidth;
+
+		const auto byteShift = bits / UINT8_WIDTH;
+		const auto bitShift = bits % UINT8_WIDTH;
+
+		std::array<uint8_t, N> result;
+
+		for (std::size_t i = 0; i < N; i++) {
+			result[i] = ((value[(i + byteShift) % N] << bitShift) | (value[(i + byteShift + 1) % N] >> (UINT8_WIDTH - bitShift))) & UINT8_MAX;
+		}
+
+		return result;
+	}
+
+	template<std::size_t N>
+	static std::array<uint8_t, N> addArray(const std::array<uint8_t, N>& a, const std::array<uint8_t, N>& b) {
+		std::array<uint8_t, N> result;
+		std::size_t sum = 0;
+		std::size_t carry = 0;
+
+		for (std::int64_t i = N - 1; i >= 0; i--) {
+			sum = a[i] + b[i] + carry;
+			carry = sum >> UINT8_WIDTH;
+			result[i] = static_cast<std::uint8_t>(sum & UINT8_MAX);
+		}
+
+		return result;
+	}
+
+	template<std::size_t N>
+	static std::array<uint8_t, N> xorArray(const std::array<uint8_t, N>& a, const std::array<uint8_t, N>& b) {
+		std::array<uint8_t, N> result;
+
+		for (std::size_t i = 0; i < N; i++) {
+			result[i] = a[i] ^ b[i];
+		}
+
+		return result;
+	}
+
+	static std::optional<AESKey> createKeyFromHex(const std::string& hex) {
+		if (hex.size() < 32) {
+			return {};
+		}
+
+		AESKey rawKey;
+		for (std::size_t i = 0; i < rawKey.size(); i++) {
+			rawKey[i] = static_cast<uint8_t>(std::stoi(hex.substr(i * 2, 2), 0, 16));
+		}
+
+		return rawKey;
+	}
+
+	struct AESKeySlot {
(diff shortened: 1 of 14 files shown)


include/crypto/aes_engine.hpp
```diff
@@ -0,0 +1,167 @@
...
+		const auto byteShift = bits / UINT8_WIDTH;
+		const auto bitShift = bits % UINT8_WIDTH;
+
+		std::array<uint8_t, N> result;
+
+		for (std::size_t i = 0; i < N; i++) {
+			result[i] = ((value[(i + byteShift) % N] << bitShift) | (value[(i + byteShift + 1) % N] >> (UINT8_WIDTH - bitShift))) & UINT8_MAX;
+		}
+
+		return result;
```
wheremyfoodat, line 32: Can the uses of UINT8_WIDTH here be replaced with `constexpr size_t u8_width = 8;` UINT8_WIDTH is an optional C23 addon

src/core/loader/ncch.cpp
```diff
@@ -1,142 +1,303 @@
...
+
+		const u64 jumpID = *(u64*)&exheader[0x1C0 + 0x8];
+
+		// It seems like some decryption tools will decrypt the file, without actually setting the NoCrypto flag in the NCCH header
+		// This is a nice and easy hack to see if a file is pretending to be encrypted, taken from 3DMoo and Citra
+		if (u32(programID) == u32(jumpID) && encrypted) {
+			printf("NCSD is supposedly ecrypted but not actually encrypted\n");
+			encrypted = false;
+		}
+		// If it's truely encrypted, we need to read section again.
```
wheremyfoodat, line 121: *truly :D

wheremyfoodat:
I don't see anything wrong and I trust you with crypto stuff but it would be nice to get a second review from someone else.
Has this been tested on multiple encrypted/unencrypted games?

wheremyfoodat:
Probably want to make the AES CTR key optionals (keyX, keyX, normalKey, encryptionInfo) explicitly initialized to std::nullopt

marysaka:
> I don't see anything wrong and I trust you with crypto stuff but it would be nice to get a second review from someone else. Has this been tested on multiple encrypted/unencrypted games?

I tested it with around 10 different encrypted games and then with decrypted form.

Wunkolo (approved):
panda

include/crypto/aes_engine.hpp
```diff
@@ -0,0 +1,167 @@
...
+#include <cstring>
+#include <cstdint>
+#include <climits>
+#include <filesystem>
+#include <optional>
+
+#include "helpers.hpp"
+
+namespace Crypto {
+	constexpr std::size_t AesKeySize = 0x10;
```
Wunkolo, line 13: `std::size_t` -> `usize`
`uint8_t` -> `u8`
etc

include/crypto/aes_engine.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+	using AESKey = std::array<uint8_t, AesKeySize>;
+
+	template<std::size_t N>
+	static std::array<uint8_t, N> rolArray(const std::array<uint8_t, N>& value, std::size_t bits) {
```
Wunkolo, line 18: All of these `const std::array<uint8_t, N>&` can be `std::span<const u8, N>` instead so it can take any container of randomly-addressable continuous elements.

src/emulator.cpp
```diff
@@ -135,8 +135,15 @@ bool Emulator::loadROM(const std::filesystem::path& path) {
     // Inside that path, we be use a game-specific folder as well. Eg if we were loading a ROM called PenguinDemo.3ds, the savedata would be in
     // %APPDATA%/Alber/PenguinDemo/SaveData on Windows, and so on. We do this because games save data in their own filesystem on the cart
     char* appData = SDL_GetPrefPath(nullptr, "Alber");
-    const std::filesystem::path dataPath = std::filesystem::path(appData) / path.filename().stem();
+    const std::filesystem::path appDataPath = std::filesystem::path(appData);
+    const std::filesystem::path dataPath = appDataPath / path.filename().stem();
+    const std::filesystem::path aesKeysPath = appDataPath / "sysdata" / "aes_keys.txt";
     IOFile::setAppDataDir(dataPath);
+
+    if (std::filesystem::exists(aesKeysPath)) {
```
Wunkolo, line 143: This can throw an exception in the case of an OS-level error. Might want to use the `std::error_code` form of this function instead to explicitly report them.
https://en.cppreference.com/w/cpp/filesystem/exists

Wunkolo (changes requested):
just some nit picks

include/crypto/aes_engine.hpp
wheremyfoodat: I changed the uint8_t to u8 but the codebase doesn't use the `usize` type in helpers.hpp that much atm.
Maybe we should have a poll among contributors about it?

include/crypto/aes_engine.hpp
wheremyfoodat: The std::span method actually is likely less intuitive here actually, I think

```cc
constexpr void updateNormalKey(std::size_t slotId) {
			if (m_generator.has_value() && hasKeyX(slotId) && hasKeyY(slotId)) {
				auto &keySlot = m_slots.at(slotId);
				AESKey keyX = keySlot.keyX.value();
				AESKey keyY = keySlot.keyY.value();

				keySlot.normalKey = rolArray(addArray(xorArray(rolArray(keyX, 2), keyY), m_generator.value()), 87);
			}
		}
		```
		The last line has a sort of snowball to process the array which works a lot better with arrays vs spans imo
		I think we should keep it, at least for now.

PR #48, opened 2023-06-20
Add OpenGL object-naming and scoping

Wunkolo:
Adds `OpenGL::setObjectLabel` and `OpenGL::DebugScope` for naming OpenGL-objects and grouping scopes of commands to help with debugging and captures.

Both support printf-style formatting and use the appropriate function-attributes and will emit warnings if the specifiers and type do not match up. `OpenGL::DebugScope` is a raii-based wrapper around `gl{Push,Pop}DebugGroup` allowing one to simple create a `OpenGL::DebugScope` object with the formatted name and it will automatically Push/Pop the scope.

qrenderdoc_01pq6Nw2Rq

This is just the utility functions, I'll let the actual naming/scoping come later.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,8 @@ endif()
 project(Alber)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
 
+option(GPU_DEBUG_INFO "Enable additional GPU debugging info" OFF)
+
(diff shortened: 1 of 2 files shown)


wheremyfoodat:
So this is like sort of intended to be plugged in when looking for a graphics bug to add additional feedback to eg Renderdoc?

wheremyfoodat (approved):

PR #50, opened 2023-06-27
Add basic controller input using the SDL2 GameController API

nadiaholmquist:
This PR adds basic controller support with a hardcoded keymap that tries to map it as close to the physical layout on the 3DS as I could.

Buttons and circle pad are mapped, touchpad and motion sensors are not supported. Controller hot-plugging should work, though only the first controller index SDL2 sees will be used.


wheremyfoodat:
Looks mostly ready. Will change some minor things and run clang-format a bit and merge.

Wunkolo (approved):

PR #56, opened 2023-07-01
Fix Clion build folders not properly being ignored

SimoneN64:
A small first contribution :)


wheremyfoodat:
image

wheremyfoodat:
preceded in #53 apparently but the literal 200 file diff made it hard to tell

PR #57, opened 2023-07-01
Add slash to CLion build folders' .gitignore line

SimoneN64:
Makes it extra :sparkles: reliable :sparkles:


wheremyfoodat:
image

PR #64, opened 2023-07-02
Separate user and dev builds

wheremyfoodat:


wheremyfoodat:
@Wunkolo Can you help with the printf format string warnings >_<

Wunkolo:
These ones?
```
/Users/runner/work/Panda3DS/Panda3DS/src/core/memory.cpp:122:15: note: in instantiation of function template specialization 'Helpers::warn<>' requested here
                                        Helpers::warn("VRAM read!\n");
                                                 ^
/Users/runner/work/Panda3DS/Panda3DS/include/helpers.hpp:56:15: note: treat the string as an argument to avoid this
                std::printf(fmt, args...);
                            ^
                            "%s", 
/Users/runner/work/Panda3DS/Panda3DS/include/helpers.hpp:37:15: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
                std::printf(fmt, args...);
                            ^~~
```
These seem to have to do with panic/warn it looks like, which pass a `const char*` which removes the attribute of knowing that the format-string is a constant compile-time string literal and not a dynamic one.
https://github.com/wheremyfoodat/Panda3DS/blob/a8bfeaf0a97d2298da0fe2a25721fc5305d71500/include/helpers.hpp#L33-L42

I would just disable that warning: `-Wformat-nonliteral` since there is no way to pass on the context of `fmt` being a literal.

wheremyfoodat:
Testing

PR #69, opened 2023-07-04
Gaming (Lights, clipping planes and more things TBD)

wheremyfoodat:


GPUCode:
Note that clip planes can cause a variety of issues when games draw close to z = 0 due to floating point inaccuracies, or at least they do in Citra, your emulator may be different.

wheremyfoodat:
Worst case scenario, if I find a game that has a bad case of too much clipping I will start toying around with epsilons

PR #72, opened 2023-07-06
Add CXI ROM support

wheremyfoodat:


wheremyfoodat:
This should also support encrypted .cxi files but that's not been tested at all so if someone tests it let me know in the Discord server

PR #73, opened 2023-07-06
[GPU] Add texture surface cache eviction

skylersaleh:
Implemented as a simple ring buffer evicting the oldest entry for now. 

Should eliminate all texture related "Surface cache full! Add emptying!" errors

### DIFF
--- a/include/renderer_gl/renderer_gl.hpp
+++ b/include/renderer_gl/renderer_gl.hpp
@@ -45,7 +45,7 @@ class Renderer {
 
 	SurfaceCache<DepthBuffer, 10> depthBufferCache;
 	SurfaceCache<ColourBuffer, 10> colourBufferCache;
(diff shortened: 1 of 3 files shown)


wheremyfoodat:
Lgtm, I will format it and make a slight change and merge

wheremyfoodat:
I am pretty bad at modulo arithmetic but this won't properly do a ring buffer, right?
It will only work if the surface cache size is a power of 2, I think?

Consider the scenario where size = 257 and the index is about to wrap.

0xffffffffu % 257 -> 0
0 % 257 -> 0 again

I will adjust it to do `surface = cache[index]; index = (index + 1) % size` instead. Should be similar perf-wise and should ring properly

wheremyfoodat (approved):

wheremyfoodat:
Before doing a proper reveal/release of the emulator it might be good to see if the size needs to be bumped too

PR #74, opened 2023-07-06
Color and depth cache updates

skylersaleh:
Additionally made the surface cache search hit for any address that lies in the surface.

This should allow multiple races to be done in Mario Kart and fixes the intro video.


wheremyfoodat:
I think this is kind of likely to get a lot of false hits especially with the texture cache, and making the colour/depth buffers ring buffers without properly implementing writing back to VRAM and stuff can cause quite a few hard-to-debug issues. Probably best to avoid this sort of "hack".

(btw I also have boost::interval for managing, comparing, etc texture/surface intervals)

PR #85, opened 2023-07-08
Initial gyroscope support

wheremyfoodat:
This makes right click + mouse swipe report as a 3DS rotation. It's not quite great yet because the proper math is horrible (currently relative rotation angle is hardcoded based on the relevant mouse motion direction vector) and mapping euler angles to mouse swipes is not a particularly fun experience, but it seems to work for games that need situational gyroscope input like Kirby Triple Deluxe.


wheremyfoodat:
image

PR #86, opened 2023-07-08
Color and depth cache updates

wheremyfoodat:
Bringing this back


wheremyfoodat:
Will mess with this a bit before getting it into master

PR #96, opened 2023-07-09
Add initial http server stuff

OFFTKP:
PANDA? PANDA.
(repeat 31 times)

### DIFF
--- a/src/httpserver.cpp
+++ b/src/httpserver.cpp
@@ -1,3 +1,4 @@
+#ifdef PANDA3DS_ENABLE_HTTP_SERVER
 #include "httpserver.hpp"
 
 #include <vector>
@@ -98,3 +99,5 @@ void HttpServer::startHttpServer() {
 	});
 	http_thread.detach();
 }
+
+#endif  // PANDA3DS_ENABLE_HTTP_SERVER
--- a/src/httpserver.cpp
+++ b/src/httpserver.cpp
@@ -1,0 +1,100 @@
+#include "httpserver.hpp"
+
+#include <vector>
+#include <map>
+#include <thread>
+#include <fstream>
+#include <string>
+
+#include "httplib.h"
+#include "services/hid.hpp"
+
+u32 stringToKey(const std::string& key_name) {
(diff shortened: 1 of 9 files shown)


OFFTKP:
lmk what you think, I'll add another commit soon

OFFTKP:
whoopsie

wheremyfoodat:
This is starting to get too big and should be moved to a dedicated file instead of emulator.cpp

OFFTKP:
> This is starting to get too big and should be moved to a dedicated file instead of emulator.cpp

I dreaded this comment

Let's see if it compiles fine and I'll move it in an upcoming commit

Wunkolo:
Also I'd rather `Emulator::screenshot` just called a member-function to the rendering backend than use OpenGL directly.

wheremyfoodat:
Good idea, agreed

src/httpserver.cpp
```diff
@@ -0,0 +1,103 @@
...
+		{"Left", Keys::Left},
+		{"Up", Keys::Up},
+		{"Down", Keys::Down},
+		{"R", Keys::R},
+		{"L", Keys::L},
+		{"X", Keys::X},
+		{"Y", Keys::Y},
+		{"CirclePadRight", Keys::CirclePadRight},
+		{"CirclePadLeft", Keys::CirclePadLeft},
+		{"CirclePadUp", Keys::CirclePadUp},
```
skylersaleh, line 30: The SkyEmu protocol supports floating point values as inputs, so maybe CirclePadX and CirclePadY might make more sense than, up/down/left/right.

OFFTKP: I'll change it in upcoming patch

wheremyfoodat (approved):

georgemoralis:
panda!

wheremyfoodat:
panda?

georgemoralis:
ΜΟΝΟ ΠΑΝΤΑ ΤΡΕΛΑ ΚΑΙ ΚΟΡΔΕΛΑ

PR #100, opened 2023-07-10
config: Use `error_code` prototype of `filesystem::exists`

Wunkolo:
The non-error-code version of these functions are susceptible to throwing an exception in the case of system errors like permission issues or underlying device errors.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -11,7 +11,8 @@
 
 void EmulatorConfig::load(const std::filesystem::path& path) {
 	// If the configuration file does not exist, create it and return
-	if (!std::filesystem::exists(path)) {
+	std::error_code error;
+	if (!std::filesystem::exists(path, error)) {
 		save(path);
 		return;
 	}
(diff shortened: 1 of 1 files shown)


src/config.cpp
```diff
@@ -38,13 +39,17 @@ void EmulatorConfig::load(const std::filesystem::path& path) {
...
+	std::error_code error;
+	if (std::filesystem::exists(path, error)) {
 		try {
 			data = toml::parse<toml::preserve_comments>(path);
 		} catch (std::exception& ex) {
 			Helpers::warn("Got exception trying to save config file. Exception: %s\n", ex.what());
 			return;
 		}
+	} else {
+		Helpers::warn("Filesystem error accessing %s: Exception: %s\n", path.string().c_str(), error.message().c_str());
```
wheremyfoodat, line 51: Hmm, if this code path is reached it doesn't mean there was necessarily an error, does it? Only that the file doesn't exist which I don't think qualifies as an error?

I think it should be like

```cc
Helpers::warn("Filesystem error accessing %s", path.string().c_str());
if (error) {
    Helpers::warn("Exception: %s", ...);
}
```

wheremyfoodat: Other than that it's good, tell me if you want to fix it or if you prefer me doing it instead on Discord

Wunkolo: I can fix it, it def shouldn't be an error if the file just doesn't exists. I'll have it print out an error code if there was an error while checking if the file exists or not, and print out that it is creating a new configuration file in the case that it doesn't.

wheremyfoodat (approved):

PR #103, opened 2023-07-10
Better screen support, partially implement texcopy/display transfer, implement setBufferSwap and framebuffer regs

GPUCode:
Also added debug names to textures similar to Citra so they can be much more easily compared.
Screens should be less hacky now but it needs texture cache improvements to improve it further.
While at it, I've also converted the command processor and gpu physical memory access to use spans instead of raw pointers (taken from my citra cmdproc clenaup branch).


GPUCode:
Also fixes a bug where command list processing address translation didn't go through the GSP memory space (confirmed through GSP dissamebly)

PR #105, opened 2023-07-11
Move `stb_image_write` into its own translation unit

Wunkolo:
Rather than having the entire implementation within `emulator.cpp`, causing incremental builds to be much slower, give it its own translation unit `stb_image_write.c`.


wheremyfoodat:
Have we verified this code is capable of writing panda PNGs

PR #115, opened 2023-07-14
kernel: Delete object properly in deleteObjectData

merryhime:
`delete`-ing a `void*` is undefined behaviour.

Alternatively you can define a base class with a virtual destructor for the payload types.


wheremyfoodat:
Looks good to me

wheremyfoodat:
Actually, the * in the getData calls should be wrong. getData<Mutex*> for example should return a Mutex** instead of a Mutex*.

wheremyfoodat:
Err oesn't removing the top line break resetting? Threads/dummy should just break instead of panic. The data for threads is not null but stack data, which is why it was previously special cased.

PR #119, opened 2023-07-16
Approximate VFP timings

wheremyfoodat:


BlurrySquire:
What is VFP?

wheremyfoodat:
> What is VFP?

https://developer.arm.com/documentation/den0013/latest/Floating-Point/Floating-point-basics-and-the-IEEE-754-standard/ARM-VFP

PR #120, opened 2023-07-17
Add a `null` rendering backend

Wunkolo:
This adds the null rendering backend, which just stubs all of the implementation details of the rendering backend, allowing for "headless" emulation of 3DS roms.
This also adds support for multiple rendering backends, based on an index within the user's `config.toml`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,7 +95,8 @@ endif()
 
 set(SOURCE_FILES src/main.cpp src/emulator.cpp src/io_file.cpp src/config.cpp
 				 src/core/CPU/cpu_dynarmic.cpp src/core/CPU/dynarmic_cycles.cpp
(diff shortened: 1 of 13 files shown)


wheremyfoodat:
Sir I tried this and I can't see the Pokemans, I think you have a bug, please fix

skylersaleh:
> Sir I tried this and I can't see the Pokemans, I think you have a bug, please fix

Wdym, its playing Pokemon Black perfectly.

wheremyfoodat:
> 

Reshiram pic or lying

wheremyfoodat:
CI seems to be dying to IWYU stuff (no <optional> in renderer.hpp)
Also likely best to move the renderer text name functions as static functions in the `Renderer` class.

So you'd do `Renderer::typeFromString` and it would be self-explanatory without global baggage.

Wunkolo:
Addressed CI issues 👍

PR #123, opened 2023-07-18
gsp_gpu: Convert vaddr to paddr from GSP service

GPUCode:
Small fix extracted from #103


wheremyfoodat:
This is correct but will likely break due to a bug regarding the shared font. Will merge when that's better.

PR #125, opened 2023-07-20
Initialize Vulkan Context

Wunkolo:
panda

### DIFF
--- a/include/renderer_vk/renderer_vk.hpp
+++ b/include/renderer_vk/renderer_vk.hpp
@@ -23,9 +23,11 @@ class RendererVK final : public Renderer {
 
 	vk::UniqueDevice device = {};
 
+	vk::UniqueSemaphore presetWaitSemaphore = {};
+	vk::UniqueSemaphore renderDoneSemaphore = {};
 
 	vk::UniqueSwapchainKHR swapchain = {};
-	std::vector<vk::UniqueImage> swapchainImages = {};
+	std::vector<vk::Image> swapchainImages = {};
 	std::vector<vk::UniqueImageView> swapchainImageViews = {};
 
   public:
--- a/include/renderer_vk/renderer_vk.hpp
+++ b/include/renderer_vk/renderer_vk.hpp
@@ -8,6 +8,8 @@ class RendererVK final : public Renderer {
 	vk::PhysicalDevice physicalDevice = {};
 	vk::UniqueDevice device = {};
 
+	vk::UniqueSurfaceKHR surface = {};
+
 	vk::UniqueDebugUtilsMessengerEXT debugMessenger;
 
   public:
--- a/include/renderer_vk/renderer_vk.hpp
+++ b/include/renderer_vk/renderer_vk.hpp
@@ -4,17 +4,13 @@
 class GPU;
 
 class RendererVK final : public Renderer {
+	// The order of these `Unique*` members is important, they will be destroyed in RAII order
 	vk::UniqueInstance instance = {};
-	vk::PhysicalDevice physicalDevice = {};
-	vk::UniqueDevice device = {};
+	vk::UniqueDebugUtilsMessengerEXT debugMessenger = {};
 
 	vk::UniqueSurfaceKHR surface = {};
 
-	vk::UniqueSwapchainKHR swapchain = {};
-	std::vector<vk::UniqueImage> swapchainImages = {};
-	std::vector<vk::UniqueImageView> swapchainImageViews = {};
-
-	vk::UniqueDebugUtilsMessengerEXT debugMessenger = {};
+	vk::PhysicalDevice physicalDevice = {};
 
 	vk::Queue presentQueue = {};
 	u32 presentQueueFamily = ~0u;
@@ -24,6 +20,13 @@ class RendererVK final : public Renderer {
 	u32 computeQueueFamily = ~0u;
 	vk::Queue transferQueue = {};
 	u32 transferQueueFamily = ~0u;
+
+	vk::UniqueDevice device = {};
+
+
+	vk::UniqueSwapchainKHR swapchain = {};
+	std::vector<vk::UniqueImage> swapchainImages = {};
+	std::vector<vk::UniqueImageView> swapchainImageViews = {};
 
   public:
 	RendererVK(GPU& gpu, const std::array<u32, regNum>& internalRegs);
--- a/include/renderer_vk/renderer_vk.hpp
+++ b/include/renderer_vk/renderer_vk.hpp
@@ -30,6 +30,9 @@ class RendererVK final : public Renderer {
 	std::vector<vk::Image> swapchainImages = {};
 	std::vector<vk::UniqueImageView> swapchainImageViews = {};
 
+	vk::UniqueCommandPool commandPool = {};
+	vk::UniqueCommandBuffer presentCommandBuffer = {};
+
   public:
 	RendererVK(GPU& gpu, const std::array<u32, regNum>& internalRegs);
 	~RendererVK() override;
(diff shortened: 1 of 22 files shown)


georgemoralis:
panda greatness!

wheremyfoodat:
So good

wheremyfoodat:
panda

liuk7071:
panda

GPUCode (approved):
Awesome work!

wheremyfoodat:
Rerunning MacOS job to generate a build log and hopefully fix the App Bundle creation

wheremyfoodat:
Praise be to Caleb

Wunkolo:
Ready for review, further work is outside of the scope of simple initialization(swapchain robustness, mesh double/triple-buffering, etc).

include/renderer_vk/renderer_vk.hpp
```diff
@@ -0,0 +1,57 @@
+#include "renderer.hpp"
+#include "vulkan_api.hpp"
+
+class GPU;
+
+class RendererVK final : public Renderer {
+	SDL_Window* targetWindow;
+
+	// The order of these `Unique*` members is important, they will be destroyed in RAII order
+	vk::UniqueInstance instance = {};
```
wheremyfoodat, line 10: I will try seeing if I can assert this at compile time via static_assert + offsetof so nobody accidentally breaks destructor order.

src/core/renderer_vk/vk_debug.cpp
```diff
@@ -0,0 +1,156 @@
...
+		labelInfo.pLabelName = objectName.get();
+		labelInfo.color[0] = color[0];
+		labelInfo.color[1] = color[1];
+		labelInfo.color[2] = color[2];
+		labelInfo.color[3] = color[3];
+
+		commandBuffer.beginDebugUtilsLabelEXT(labelInfo);
+	}
+
+	void insertDebugLabel(vk::CommandBuffer commandBuffer, std::span<const float, 4> color, const char* format, ...) {
```
wheremyfoodat, line 127: Ditto (also this seems DRYable but eh)

include/renderer_vk/vk_debug.hpp
```diff
@@ -0,0 +1,48 @@
...
+namespace Vulkan {
+
+	VKAPI_ATTR VkBool32 VKAPI_CALL debugMessageCallback(
+		VkDebugUtilsMessageSeverityFlagBitsEXT messageSeverity, VkDebugUtilsMessageTypeFlagsEXT messageType,
+		const VkDebugUtilsMessengerCallbackDataEXT* callbackData, void* userData
+	);
+
+	void setObjectName(vk::Device device, vk::ObjectType objectType, const void* objectHandle, const char* format, ...);
+
+	template <typename T, typename = std::enable_if_t<vk::isVulkanHandleType<T>::value == true>, typename... ArgsT>
```
wheremyfoodat, line 18: This signature is legitimately fascinating to look qt

src/core/renderer_vk/renderer_vk.cpp
```diff
@@ -0,0 +1,546 @@
...
+
+	if (auto createResult = vk::createInstanceUnique(instanceInfo); createResult.result == vk::Result::eSuccess) {
+		instance = std::move(createResult.value);
+	} else {
+		Helpers::panic("Error creating Vulkan instance: %s\n", vk::to_string(createResult.result).c_str());
+	}
+	// Initialize instance-specific function pointers
+	VULKAN_HPP_DEFAULT_DISPATCHER.init(instance.get());
+
+	// Enable debug messenger if the instance was able to be created with debug_utils
```
wheremyfoodat, line 391: Vk and OpenGL debug messages should also maybe have their own CMake flag eventually but this should def stay enabled for Vk at the moment

src/core/renderer_vk/renderer_vk.cpp
```diff
@@ -0,0 +1,546 @@
...
+	if (std::find(
+			instanceExtensions.begin(), instanceExtensions.end(),
+			// std::string_view has a way to compare itself to `const char*`
+			// so by casting it, we get the actual string comparisons
+			// and not pointer-comparisons
+			std::string_view(VK_EXT_DEBUG_UTILS_EXTENSION_NAME)
+		) != instanceExtensions.end()) {
+		vk::DebugUtilsMessengerCreateInfoEXT debugCreateInfo{};
+		debugCreateInfo.messageSeverity = vk::DebugUtilsMessageSeverityFlagBitsEXT::eVerbose | vk::DebugUtilsMessageSeverityFlagBitsEXT::eInfo |
```
wheremyfoodat, line 400: The mask for verbose being so verbose makes me chuckle

src/core/renderer_vk/renderer_vk.cpp
```diff
@@ -0,0 +1,546 @@
...
+					supportResult.result == vk::Result::eSuccess) {
+					return supportResult.value;
+				}
+			}
+			return false;
+		};
+
+		partitionEnd = std::stable_partition(physicalDevices.begin(), partitionEnd, surfaceSupport);
+
+		// Prefer Discrete GPUs
```
wheremyfoodat, line 438: Unlike OpenGL, Vk lets us control device creation very well, so maybe we could add a confic about whether to prefer discrete vs integrated in the future (laptop users who don't want their laptop to run warm or eat up battery would like it maybe)

wheremyfoodat (commented):
Looks good to me, will only do very tiny neglible changes at best

include/renderer_vk/renderer_vk.hpp
Wunkolo: This may not be necessary to maintain since Vulkan validation layers will already yell at you if you destroy things in the wrong order. Like destroying an instance or device or command-pool while dependent objects still exist and such.

src/core/renderer_vk/vk_debug.cpp
```diff
@@ -0,0 +1,156 @@
...
+		nameInfo.objectType = objectType;
+		nameInfo.objectHandle = reinterpret_cast<std::uintptr_t>(objectHandle);
+		nameInfo.pObjectName = objectName.get();
+
+		if (device.setDebugUtilsObjectNameEXT(nameInfo) != vk::Result::eSuccess) {
+			// Failed to set object name
+		}
+	}
+
+	void beginDebugLabel(vk::CommandBuffer commandBuffer, std::span<const float, 4> color, const char* format, ...) {
```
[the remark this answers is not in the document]
Wunkolo: Yeah that should be fine, so we can cull this functionality out in the case of a user-facing build like we do on the OpenGL renderer.

include/renderer_vk/vk_debug.hpp
Wunkolo: It's basically a compile-time way to derive a `vk::ObjectType`-enum based on the `vk::blah` that gets passed in, and will yell at you at compile-time if you try to give it something weird.

src/core/renderer_vk/renderer_vk.cpp
Wunkolo: I def agree, this is what the driver will use to for its own diagnostic messaging as well. The NVidia driver is known to push messages through here at times in some cases as well during some crashes or exceptional moments.

src/core/renderer_vk/renderer_vk.cpp
Wunkolo: The next step I had in mind with all this "stable_partition" stuff is to have a bit of config-syntax to define the device-preference. So it will order the device-list partitioning/sorting based on a passed-in list of criteria like `devicePriority = "discrete>apiVersion>vram>"`, or whatever that config-syntax would be, to avoid having to have a whole potentially unstable device UUID defined in a config file.
The more share-able and stand-alone config files are the better.

wheremyfoodat (approved):

PR #129, opened 2023-07-20
Add preliminary cheat support

wheremyfoodat:
Add basic framework for cheats + support for some simpler ActionReplay (CTRPF) cheat codes.
Note: Some AR cheats rely on things such as linear heap mapping being 1-1 with retail kernel and as such don't work. Also cheats are not yet hooked to the GUI.


wheremyfoodat:
Disclaimer: We at Panda3DS HQ do not endorse spousal infidelity, academic cheating or cheating in multiplayer games. Thank you for your time.

PR #136, opened 2023-07-24
Refactor http server

OFFTKP:
Improvements:

- Remove Emulator::pollHttpServer function as it did not belong inside Emulator, instead give HttpServer a pointer to emulator
- Create an HttpAction class which each type of action (more to be added) derives
- Instead of a single request, can now queue many requests in a (hopefully) thread safe queue and execute them whenever polled
- Use RAII for starting and stopping the server
- Cleaner HttpServer class overall
- Added function that checks if a button is pressed in hid.hpp


wheremyfoodat:
This is a weird thing to explain but I'll be replacing the HID isPressed function you added for New3DS compatibility reasons

The tl;dr is that due to the way the New3DS extra buttons are implemented, this will not be compatible with ZL/ZR/C stick

OFFTKP:
That's fine

wheremyfoodat (approved):

PR #139, opened 2023-07-26
Add load_rom http server command

OFFTKP:
Panda.


liuk7071:
Panda.

wheremyfoodat:
We really need some way to report if ROM loading was actually successful. This will work for eg GBA ROMs which barely have a defined format to begin with, but less so for 3DS ROMs which have relatively complicated and strict formats.

wheremyfoodat:
So I guess this will also need the deferred response API

PR #142, opened 2023-07-27
[Vulkan] Implement framebuffer management

Wunkolo:
Some initial framebuffer management. Panda3DS will maintain a cache of its own framebuffer objects and will blit them to the swapchain-image rather than operating on the swapchain image directly.


wheremyfoodat:
C++20 ranges (alongside concepts) will have to be avoided in the codebase for now unless necessary because AppleClang is stuck way behind LLVM Clang, and that's the one most MacOS people have readily available (Apple makes using real Clang & GCC unnecessarily hard in an attempt to push AppleClang)

Wunkolo:
AppleClang is rly holding the whole Apple software ecosystem back 😒. I'll revert the ranges stuff.

wheremyfoodat:
My favourite part about MacOS is that invoking `gcc` actually invokes AppleClang

wheremyfoodat:
@Wunkolo Would it be easier for you to do an initial merge so that you can avoid getting crippled by merge conflicts and such? It's alright if you'd like that, this rendition of the code might not be ready for games but it's still better than upstream ^_^

Wunkolo:
That might be worth doing yeah. I'm like at the last 15% of this feature before it's minimally viable and for cleanups to start happening with later PRs, then more PRs for actually rendering after.
I'll mark it as ready-for-review once it's ready for that so I can break up the tasks rather than have huge PRs like this one.

wheremyfoodat:
OK drop a message here when you want it merged

Wunkolo:
In a good state for now. Will have to revisit a lot of these nasty patterns later with some further, more broken-up, PRs but this is a draft implementation for all the things needed to get some basic render-textures working(allocating, displaying, transposing within a shader, clearing, etc).

Runs with no validation errors when running both_screens.elf, SimplerTri.elf, and textured_cube.elf.

While testing some retail games while working on this:
On my Intel GPU(UHD Graphics 750): Runs many retail games with no validation layer messages.
On my AMD GPU(RX 560D): Same as above.
On my Nvidia GPU(RTX 3090): Some games use RGBA4 render-textureswWhich is mapped to the `R4G4B4A4_UNORM_PACK16` type in Vulkan at the moment).
Nvidia only supports this format as a blit-source and not for render-pass attachments or blitting _into_. This will need some runtime copying back and forth into something like `R8G8B8A8_UNORM`.

wheremyfoodat (approved):

wheremyfoodat:
Pamda

PR #155, opened 2023-08-03
gpu: Better screen support (and other things to pander to the panders)

GPUCode:
Same as the previous PR but trimmed to the essentials. ~~Fb rectangles with be in the next PR~~.


src/core/services/gsp_gpu.cpp
```diff
@@ -19,16 +21,14 @@ namespace ServiceCommands {
 }
 
 // Commands written to shared memory and processed by TriggerCmdReqQueue
-namespace GXCommands {
```
wheremyfoodat, line 22: See previous

src/core/PICA/regs.cpp
```diff
@@ -162,7 +187,7 @@ void GPU::writeInternalReg(u32 index, u32 value, u32 mask) {
 			}
 			break;
 
-		// Restart immediate mode primitive drawing
+			// Restart immediate mode primitive drawing
```
wheremyfoodat, line 190: Sir your Qt creator is broken

include/renderer_gl/renderer_gl.hpp
```diff
@@ -44,8 +44,8 @@ class RendererGL final : public Renderer {
 	float oldDepthOffset = 0.0;
 	bool oldDepthmapEnable = false;
 
-	SurfaceCache<DepthBuffer, 10, true> depthBufferCache;
```
wheremyfoodat, line 47: Is there a particular use case that inspired going from 10 to 64?

include/renderer_gl/surfaces.hpp
```diff
@@ -29,12 +33,32 @@ struct ColourBuffer {
...
+		fmt = formats[(int)format];
+		type = types[(int)format];
+
+
         // Create texture for the FBO, setting up filters and the like
         // Reading back the current texture is slow, but allocate calls should be few and far between.
         // If this becomes a bottleneck, we can fix it semi-easily
         auto prevTexture = OpenGL::getTex2D();
-        texture.create(size.x(), size.y(), GL_RGBA8);
+        texture.create(size.x(), size.y(), internalFormat);
```
wheremyfoodat, line 61: Won't this end up degrading image quality? I initially used GL_RGBA8 here always hoping that the extended precision would make for nicer textures.

GPUCode: It's a question if you want to be more accurate to the hardware or better image quality. I haven't really tested if this affects image quality, though having the correct format makes texture downloads simpler. Also in vulkan if you have the correct format you can do reinterpretation with a simple image copy instead of shader in some cases.

include/renderer_gl/renderer_gl.hpp
GPUCode: Had some crashes and bumping the number makes games more stable.

wheremyfoodat: Try configuring CMake with -DDISABLE_PANIC_DEV=ON
This will make these games not crash and use a ring buffer instead

wheremyfoodat: (Prolly need to delete CMakeCache first)

GPUCode: Should I keep the bumped number or revert it back to 10?

wheremyfoodat: I'll bump it down for now and consider more carefully how to adjust it in the guture should the need arise

ArbestRiAgain:
Fixes the stretched bottom screen in 3D Land and many other games.
image

include/renderer_gl/opengl.hpp
```diff
@@ -615,83 +615,4 @@ namespace OpenGL {
         glBlendFuncSeparate(fac1, fac2, fac3, fac4);
     }
 
-    // Abstraction for GLSL vectors
-    template <typename T, int size>
```
wheremyfoodat, line 619: There was no way for you to know this lol but I'll bring these back. Reason is that this header is actually a standalone library several other projects use so I need to be able to drag and drop the latest Panda version to those projects. This means I can't make breaking ABI changes like this

GPUCode: The header should be in externals then

wheremyfoodat: True imma move it when I'm on the keeb again

wheremyfoodat:
Last commit fixes an "issue" where the compiler straight up does not optimize `gpu.readExternalRegs(Framebuffer0Addr)` and emits a whole load GPU object + call readExternalRegs + bounds check + read + return. This is actually impossible to optimize without aggressive LTO (and that is still an if).

Now it should be closer to a regular variable access instead.

I am also not sure if my relocating externalRegs for the reason I mentioned is 100% a good idea. My intuition says so, but I may reconsider in the future.

include/renderer_gl/surfaces.hpp
```diff
@@ -79,7 +113,7 @@ struct ColourBuffer {
 struct DepthBuffer {
     u32 location;
     PICA::DepthFmt format;
-    OpenGL::uvec2 size; // Implicitly set to the size of the framebuffer
+    Math::uvec2 size; // Implicitly set to the size of the framebuffer
```
wheremyfoodat, line 116: Nowadays we have glm too btw for all vector/matrix/quaternion needs but the "old" vec implementation is fine here too since there's no math being really done.

(This is just an informational comment, nothing needs to change here)

src/core/services/gsp_gpu.cpp
```diff
@@ -261,6 +261,18 @@ void GPUService::setAxiConfigQoSMode(u32 messagePointer) {
 	mem.write32(messagePointer + 4, Result::Success);
 }
 
+void GPUService::setBufferSwap(u32 messagePointer) {
```
wheremyfoodat, line 264: This would make games crash via err:f since it does not handle the IPC response header or the IPC result code. I am not sure what game to use to test it though :/

If there's none we know of atm, I will keep the code but add a Helpers::panic to revisit this when a game like that is found.

src/core/renderer_gl/renderer_gl.cpp
```diff
@@ -550,42 +579,64 @@ OpenGL::Texture RendererGL::getTexture(Texture& tex) {
...
-	const u32 outputGap = outputSize >> 16;
+	u32 outputWidth = outputSize & 0xffff;
+	u32 outputHeight = outputSize >> 16;
 
-	auto framebuffer = colourBufferCache.findFromAddress(inputAddr);
-	// If there's a framebuffer at this address, use it. Otherwise go back to our old hack and display framebuffer 0
-	// Displays are hard I really don't want to try implementing them because getting a fast solution is terrible
-	OpenGL::Texture& tex = framebuffer.has_value() ? framebuffer.value().get().texture : colourBufferCache[0].texture;
+	if (inputWidth != outputWidth) {
+		// Helpers::warn("Strided display transfer is not handled correctly!\n");
```
GPUCode, line 602: Why did you comment this out? Games may silently break due to that

wheremyfoodat: FE: A has it trigger too much to the point of making it crawl.
I can lock it behind an

```cc
if constexpr (!Helpers::isUserBuild()) {}
```

GPUCode:
Majoras Mask is broken because of strided copy. I'll add it to this PR

wheremyfoodat:
OK, leave a message here when I can continue

src/core/services/gsp_gpu.cpp
wheremyfoodat: Fixed the IPC buffer for SetBufferSwap, still added a panic for when there's a known test case

GPUCode:
Added basic texture copy implementation that should work in the majority of cases. I've also labeled textures and added debug scopes for display transfer and texture copy so debugging in renderdoc is made easier
εικόνα

src/core/renderer_gl/renderer_gl.cpp
```diff
@@ -616,18 +615,88 @@ void RendererGL::displayTransfer(u32 inputAddr, u32 outputAddr, u32 inputSize, u
...
+	Math::Rect<u32> destRect = destFramebuffer->getSubRect(outputAddr, outputWidth, outputHeight);
+
+	if (inputWidth != outputWidth) {
+		// Helpers::warn("Strided display transfer is not handled correctly!\n");
+	}
 
 	// Blit the framebuffers
-	srcFramebuffer.fbo.bind(OpenGL::ReadFramebuffer);
-	destFramebuffer.fbo.bind(OpenGL::DrawFramebuffer);
+	srcFramebuffer->fbo.bind(OpenGL::ReadFramebuffer);
```
wheremyfoodat, line 625: Unchecked optional access?

GPUCode: optional only matters if createIfNotFound is false, otherwise the function is guarantted to return a surface

wheremyfoodat:
Threw the format calls for OpenGL debug behind `#ifdef GPU_DEBUG_INFO` because they are impossible to optimize out otherwise.
I will look to see why NSMB2 is not a huge fan of the textureCopy implementation

GPUCode:
Latest commit fixes jitter in MK7 bottom screen

wheremyfoodat:
Resolving conflicts in a bit

wheremyfoodat:
Should be resolved

wheremyfoodat:
More conflicts
256956941-cf3ef8ad-5b68-4885-a0e2-67651f20ee3a

GPUCode:
Latest commit fixes Band of Thieves and 1000 Pokemon and also corrects the initial splash screen in Captain Toad
εικόνα

wheremyfoodat:
Omg
> Band of Thieves and 1000 Pokemon

Sir you're just making up games now

wheremyfoodat:
Any game to test v flip with?

GPUCode:
> Any game to test v flip with?

Don't remember any off the top of my head, I added it because it was quite trivial

src/core/renderer_gl/renderer_gl.cpp
```diff
@@ -473,6 +473,7 @@ void RendererGL::display() {
 	auto bottomScreen = colourBufferCache.findFromAddress(bottomScreenAddr);
 	screenFramebuffer.bind(OpenGL::DrawFramebuffer);
 
+	OpenGL::setClearColor(0.f);
```
wheremyfoodat, line 476: I will add clear colour/depth/stencil to gl_state because might as well

include/renderer_gl/surfaces.hpp
wheremyfoodat: Reverting temporarily

wheremyfoodat (approved):

wheremyfoodat:
image
image

PR #158, opened 2023-08-05
Initial work on HLE applets

wheremyfoodat:
Aims to fix games relying on mii selector and potentially sw keyboard
Currently breaks them all though (oops)


wheremyfoodat:
This PR is getting split into 2 parts and the 1st part is being merged now for scope creep reasons.

PR #164, opened 2023-08-09
HLE improvements for Chibi-Robo et al

wheremyfoodat:
Currently Chibi-Robo crashes on a null access so need more work before merging, might be a thread issue >_<


wheremyfoodat:
Crash seems unrelated, bleh

PR #170, opened 2023-08-10
Add soc:u

wheremyfoodat:
Gets miitopia further, sadly game throws a fatal error after a few seconds because it needs an SD card (Can reproduce the crash on Citra if I tell it to not use an SD)


wheremyfoodat:
# Panda

PR #173, opened 2023-08-11
More efficient updating of TEV state

wheremyfoodat:
Moves TEV state parsing to vertex shader to avoid 5 uniform uplodads per draw call + some memory moves


wheremyfoodat:
Thoughts (and prayers appreciated) @Wunkolo @GPUCode @fleroviux @skylersaleh

Wunkolo (approved):
LGTM!

fleroviux:
This seems like the logical thing to do now that we have access to all PICA IO in the shader. LGTM!

wheremyfoodat:
re: Discord

Will switch to reading in the fragment shade instead to avoid going over input/output attribute caps

PR #179, opened 2023-08-13
Autodetect 3DS region and properly report it in CFG::SecureInfoGetRegion

wheremyfoodat:
This fixes compatibility issues with various games that have internal region lock, such as Luigi's Mansion or Rhythm Heaven Paradise/Megamix


wheremyfoodat:
This fixes a lot of reported problems with non-USA dumps, since the previous implementation would always report the region as USA

PR #183, opened 2023-08-14
Render to texture fixes

GPUCode:
This addresses a couple of current issues when games perform render to texture tricks.

* When binding texture units we now check the colour buffer cache first, in case the game is trying to use a framebuffer
  as texture source. This is not an ideal solution; the proper way would be to merge the texture and colour buffer caches but
  that requires a more thought out invalidation scheme and this approach works for the majority of cases.

* At the moment textures are loaded normally and the texture coordinates are flipped in the vertex shader to account for the difference between opengl and pica texture coordinate origin. However this presents problems as the afformentioned framebuffers are not in pica origin, producing incorrect results. To address this textures are loaded flipped instead of flipping coordinates. With this Samus Returns render to texture effect is correct, position wise and the top screen is now blue
samus

* Bumps colour buffer cache size to 32 as certain games already exceed the current limit, resulting in flashes. Iterating over an std::array is very fast so I don't expect this to result in any performance loss. Combined with above fix FE:A renders properly

awake

### DIFF
--- a/include/renderer_gl/renderer_gl.hpp
+++ b/include/renderer_gl/renderer_gl.hpp
@@ -45,7 +45,7 @@ class RendererGL final : public Renderer {
 	bool oldDepthmapEnable = false;
 
 	SurfaceCache<DepthBuffer, 16, true> depthBufferCache;
-	SurfaceCache<ColourBuffer, 16, true> colourBufferCache;
+	SurfaceCache<ColourBuffer, 32, true> colourBufferCache;
 	SurfaceCache<Texture, 256, true> textureCache;
 
 	// Dummy VAO/VBO for blitting the final output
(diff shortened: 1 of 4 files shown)


wheremyfoodat:
Not the biggest fan of the up-to-100-iterations linear scan on every draw call

GPUCode:
My arguments in favour of this:
* Linear iteration is one of the fastest operations on a collection, especially on an array. 100 elements is not a lot on the grand scheme of things.
* It's beneficial as it fixes a bunch of cases where this is a problem
* Microoptimization in such early stages may slow down progress. If this turns out a problem it can be reverted at any time.

wheremyfoodat:
Linear iteration is pretty terrible especially when using a relatively complex comparator function on non-trivial data in AoS formation, and when assuming 10 draw calls per frame can lead too 60 thousand useless searches per second.

It's less so micro-optimization, moreso avoiding digging myself a grave to dig myself out of later

GPUCode:
From my testing even iterating 1.000.000+ AoS takes less than 300 ns https://godbolt.org/z/a18GW53G1
Regardless unless the alternative is implemented I'll need to close this PR as all the other fixes depend on that one. It's your call.

wheremyfoodat:
The texture flip part seems nice

PR #189, opened 2023-08-15
Stub Y2R::SetPackageParameter, upstream the positional light change

wheremyfoodat:


wheremyfoodat:
Fixes some Atlus games like Persona Q and SMT: Devil Summoner Soul Hackers

PR #215, opened 2023-08-19
Stub update RomFS

wheremyfoodat:
Since we currently don't support loading game updates or DLC, reading the update RomFS just reads the cartridge RomFS.

When we do decide to support updates/DLC, we will have to implement overlaying the update filesystems with the ROM filesystems (RomFS/ExeFS) if an update/DLC is loaded


wheremyfoodat:
Gets Shantae playable
image

PR #216, opened 2023-08-20
Stub DLC::GetPatchTitleInfo to fail

wheremyfoodat:


wheremyfoodat:
Also added AM::RegisterDisconnectEvent, hit me up if you want to Netflix n Chill
image

PR #217, opened 2023-08-20
Add the GetThreadIdealProcessor SVC and fix UB

wheremyfoodat:


src/core/loader/ncch.cpp
```diff
@@ -286,6 +286,7 @@ bool NCCH::parseSMDH(const std::vector<u8>& smdh) {
 	} else if (taiwan) {
 		region = Regions::Taiwan;
 	}
+	return true;
```
Wunkolo, line 289: The weirdest thing. I sometimes hit an illegal instruction around here for a bit. If that is the UB that this was addressing.

wheremyfoodat: If it's a recent thing then probaby. This was added in #179

PR #220, opened 2023-08-20
Trying MacOS 10.15

wheremyfoodat:


wheremyfoodat:
On hold at the moment, 10.15 app bundle seems to not work properly

PR #244, opened 2023-08-29
gpu: Saturate color before interpolation

GPUCode:
Fixes the stars in Cut the Rope. std::min cannot be used because color may be NaN
εικόνα


wheremyfoodat:
:rhapp:
Can't this be better done in vertex shader though?

GPUCode:
Done

wheremyfoodat (approved):

wheremyfoodat:
I will probably be asleep soon so I will give you perms to merge it.
Also I will have to make a label for this sort of PR cause we'll need to pay attention to this edge case when we do sw rendering

GPUCode:
Peach going to sleep early whaa :open_mouth:

PR #251, opened 2023-09-02
added feature: 3dsx loading

LiquidFenrir:
3DSX-embedded romFS works (tested with the gpusprites and cubemap example from 3ds-examples)
Haven't tested if this affected other file formats, it shouldn't.
Maybe the cxi romfs loading, since that was moved into a if statement, but I heavily doubt it, Still a good idea to test if you can :)

Pretty much entirely adapted from Luma3DS's loader reimpl 3dsx loading code, thanks @TuxSH for that bit of GPLv3 wisdom


wheremyfoodat:
So good

wheremyfoodat:
Thank you for your contributions and for being panda
image

PR #272, opened 2023-09-13
Add CRO support

noumidev:
This pull request implements several ldr:ro functions.

Some stuff hasn't been implemented yet (proper CRO fixing, clearing relocations upon unloading for CROs with a fix level of 0, ...) and several games are still complaining, but this should be good enough for some testing.


wheremyfoodat (approved):

wheremyfoodat:
Already looked at the code, getting merged when CIs are over
image-29

PR #280, opened 2023-09-17
Embed LuaJIT part 1

wheremyfoodat:
image


ArbestRiAgain:
What does LuaJIT do?

OFFTKP:
Lua is a programming language that is going to be used for scripting support in Panda3DS. LuaJIT is a JIT compiler for Lua to make it go fast.

PR #288, opened 2023-09-26
Services and commands for Home Menu

noumidev:
This pull request adds a lot of commands and services required for Home Menu.

Unfortunately, Home Menu doesn't quite work yet (performs a lot of invalid memory accesses, eventually gets stuck reading from more bad addresses), but I think this is a step in the right direction.


ArbestRiAgain:
HOME MENU SUPPORT FINALLY! Well, it doesn’t get far yet…

noumidev:
https://github.com/wheremyfoodat/Panda3DS/pull/296

Issue #190, opened 2023-08-16
[Feature request] Add a Fullscreen mode to the emulator or a resizable window


LaReunion974:
Is it possible to make a Fullscreen mode or an argument when we boot a game with a command line to make them boot in Fullscreen or make the Alber window resizable because it's very small

wheremyfoodat:
That is indeed a good feature request and something that should be done

LaReunion974:
It will be a great feature ^^

wheremyfoodat:
Largely

PR #304, opened 2023-10-01
Make emulator output size properly configurable

wheremyfoodat:
Does most of the work for #190, so I'll be closing that and resizing should be implemented relatively soon in panda-qt.
panda-sdl will get resizing sometime in the future instead.


wheremyfoodat:
Also shoved in some code to split GL-based and Vk-based backends more cleanly for Qt, and added missing translations

PR #313, opened 2023-10-17
Fix more shader type mismatches

OFFTKP:


wheremyfoodat:
Don't the literals in lutLookup also need to use unsigned literals

PR #314, opened 2023-10-17
Fix android app data path

OFFTKP:
Also AndroidNdkModules didn't exist in my installation (may be an Archlinux NDK package thing), but its only included for a small macro android_ndk_import_module_cpufeatures, so I just replaced it with its definition, see https://android.googlesource.com/platform/external/android-cmake/+/refs/heads/upstream-master/AndroidNdkModules.cmake


OFFTKP:
SDL_GetPrefPath would fail on android for me with "Could not find JNI environment" or something similar, but this /data/data/<appname>/ path is the equivalent of what we want here (and I think what GetPrefPath would return anyway)

wheremyfoodat (approved):

PR #315, opened 2023-10-17
wrage?

OFFTKP:
Use a different loading function for android, otherwise it would not load some functions like glad_glTexStorage2D which normally gets loaded at load_GL_VERSION_4_2 which android was probably missing, and now loads from load_GL_ES_VERSION_3_0.

A standalone function is also helpful to further separate the emulator core from the UI


wheremyfoodat:
Needs to check the `PANDA3DS_ENABLE_OPENGL` define before defining this, also likely better to put this somewhere outside of emulator.cpp

OFFTKP:
There's some code duplication (frontend_qt code also exists in frontend_sdl). I was thinking of making the initFrontend function handle that common code in emulator.cpp and call an initFrontendInternal which would be the frontend specific code defined in frontend_sdl/qt but not sure how common the duplicated code is (would it be needed for every possible frontend panda might have in the future?)

OFFTKP:
Apart from ease to adding new frontends, I think this makes it easier to create unit tests in the future. Say you want to run the emulator class as a standalone, only initialize opengl for testing. You could add another ifdef in emulator.cpp, or you could simply define `initFrontend` (which maybe should be named initEnvironment) in the test file to initialize the environment the way you want for your testing purposes.

wheremyfoodat:
Can sdl_frontend.hpp include <filesystem> rq
Merging this right after

wheremyfoodat:
Wait you included it in the .cpp not .hpp
:exploding_head:

PR #324, opened 2023-10-21
Add icon for hydra

wheremyfoodat:


OFFTKP:
lgtm

wheremyfoodat:
But does it work in the frontend

wheremyfoodat:
Oh the rgba data might be endian flipped but it would still be good to know if the frontend shows something

OFFTKP:
hmm now that I think of it perhaps requiring the definition of a global function is better. Because then this way I need to make an emulator instance with createEmulator to get the icon.

OFFTKP:
image

wheremyfoodat:
How do you report whether there's an icon or not with this API though

PR #325, opened 2023-10-23
Upload hydra core artifacts

OFFTKP:


wheremyfoodat:
MacOS is probably broken here, I purposely made the core not create an App Bundle for the dynlib (since, well, it's a dynlib and not an app)

OFFTKP:
lmao the android core is 120MB :wrage:

PR #328, opened 2023-10-24
Make cheats struct more versatile

OFFTKP:


OFFTKP:
Fixed merge conflict and rebased

PR #337, opened 2023-11-01
Add more HLE Y2R functions

wheremyfoodat:
Used by Street Fighter IV and more.
Sadly Street Fighter IV needs some more KTimer fixes to fully go in-game, since it uses timers to time the opening FMV which doesn't work with the current timer stub.


wheremyfoodat:
Discord bot test

PR #341, opened 2023-11-18
Add cpu_kvm file

OFFTKP:


Ishan09811 (changes requested):
you need to add ``#include <cstdio>`` and
``#include <cstdint>`` for uintptr_t and printf then your compiling errors may fix

PR #345, opened 2023-12-06
Fix Kernel WaitSynchronization1 HLE, properly report CirclePadPro as not connected

wheremyfoodat:
Fixes games stuck polling IR (eg Majora's Mask 3D) and other games that may have been broken due to WaitSync timeouts not working


wheremyfoodat:
Terraria gets further but still doesn't boot >:(

PR #360, opened 2024-01-02
Implement SVC 0x3B (GetThreadContext)

wheremyfoodat:
image

K11 decomp seems to indicate that svcGetThreadContext is not a stub in the actual retail kernel, but this seems to work fine enough for now. Added a warning in case it needs to be revisited, still


wheremyfoodat:
3DBrew does say this but I'm pretty sure that's just hallucinated
image

PR #363, opened 2024-01-03
Apply GLES patch on CI

OFFTKP:


wheremyfoodat:
v8.1 -> v8 causes build on arm64 to fail due to no CRC instructions. Need to figure out some CMake crap to propagate Cryptopp's NO_ASM option to its CMakeLists
image

OFFTKP:
(or to pass +crc)

wheremyfoodat:
It's pretty likely that processors that don't have ARMv8.1 will not have the CRC extensions

PR #364, opened 2024-01-04
Fix cmp in x64 shader JIT

wheremyfoodat:
Should fix graphics in a bunch of games
image


wheremyfoodat:
Final Fantasy working if you don't need bottom left of your screen
image

PR #368, opened 2024-01-06
Update hydra submodule & add settings for hydra core

OFFTKP:


OFFTKP:
draft - need to actually get and use these settings

PR #371, opened 2024-01-18
WIP fragment pipeline GLSL recompiler

wheremyfoodat:
This PR will be adding the ability to generate GLSL fragment shaders for emulating the PICA fixed function pipeline, alongside the current ubershader-based solution.

This is important for supporting lower end GPUs, especially mobile ones, and will be useful for implementing more GPU features without significant performance loss.

Note that this will not completely replace the slower-but-stutterless ubershader solution. It will start off as an optional feature that's disabled by default, and then hopefully hybrid shader emulation will be done in order to make the emulator both fast and stutter-free.


skylersaleh:
Pretty interested to see how this performs. One aspect that we do need to do, even on the uber shader version is to split out the shader for features that put GPUs on the slow path (ie. discard and fragment-z). So that we don't operate on the slow path when those features aren't used by the draw call.

wheremyfoodat:
> Pretty interested to see how this performs. One aspect that we do need to do, even on the uber shader version is to split out the shader for features that put GPUs on the slow path (ie. discard and fragment-z). So that we don't operate on the slow path when those features aren't used by the draw call.

Yeah i wanted to make ubershaders at least slightly generated too, to generate some permutations for things like alpha test config on boot time

PR #376, opened 2024-01-23
More camera emulation

wheremyfoodat:
Properly implements camera ports and more CAM:U functions, making fewer games crash due to unimplemented camera stuff. Also adds FS:GetSdmcArchiveResource because @georgemoralis wants to play Ridge Racer

image
image


georgemoralis:
aaaaa ridge racer!

PR #381, opened 2024-01-25
Make navigation Bar Fully Transparent

Ishan09811:


wheremyfoodat:
This seems ok cc @GabrielBRDeveloper?

PR #382, opened 2024-01-25
PandaGlSurfaceView : Optimisations

Ishan09811:


wheremyfoodat:
This PR doesn't really add any optimizations, doesn't compile, breaks formatting and some of the things it does will very likely cause problems down the line (eg shutting down the emulator in SurfaceDestroy will break if the Android port adds multithreading or for some reason the emulator core wants to outlive the drawing surface)

The tiny amount of resource leakage here is also fine, since it's only 2 small GL objects and they aren't really leaked since the driver will automatically free them when the program closes, ie when they actually die.

PR #385, opened 2024-01-26
more smoother experience on selecting the navigation items

Ishan09811:


wheremyfoodat:
This is a downgrade...? Deferring the commit can lead to other problems like duplicate requests, and this transaction is very rare and cheap so there's no problem in using commitNow. It's more or less instant at the moment on my phone.

Ishan09811:
@wheremyfoodat 
commitNow() method performs the transaction immediately, which causes visual jank so the commit method is better performing and smoother experiences for users and if you have the duplicate requests issue then I can solve it.

PR #390, opened 2024-01-28
apk optimisation + compile both release and debug build 

Ishan09811:


wheremyfoodat:
Uploading the release build fails 🤔

Ishan09811:
@wheremyfoodat 
I don't know why but the release build is can't find in the directory to upload it so you can solve it ?

wheremyfoodat:
The directory doesn't exist because the APK is unsigned, I'll look if it's possible to sign it with the default debug key

wheremyfoodat:
Seems like the release APK now builds but won't open a real phone

Ishan09811:
@wheremyfoodat 
it's now uploading the release apk btw thanks for the help

wheremyfoodat:
Well yes but it doesn't open

Ishan09811:
@wheremyfoodat 
maybe because of the debug and release small characters , maybe we need to use it in capital.

Ishan09811:
@wheremyfoodat 
maybe it's causing because of the debug key

wheremyfoodat:
This doesn't correctly pass the build type to CMake https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#id37

wheremyfoodat:
OK so the cmake docs say it's case sensitive but it's also case insensitive because that's very not confusing, this probably works then

PR #391, opened 2024-01-28
renderer_vk: Better Vulkan support

GPUCode:
This is an almost complete overhaul of the existing vulkan backend, so it can now boot games. It's still not perfect and doesn't have the same compatibility as OpenGL but still a start. Vulkan version used is 1.2 as it supports timeline semaphores which make our life easier, this means that for Android support to occur a timeline semaphore emulation layers needs to be written.

Given RendererVK was already quite large from all the initialization code, it has been split up into separate classes. Instance creates the instance, logical device and the VMA allocator. Scheduler provides an easy to use interface over command buffers and manages their lifetime. RenderManager tracks the current renderpass instance, so each draw doesn't have to start and end the current pass. It also caches framebuffers and renderpass objects. PipelineManager holds all the pipelines that map to emulated PICA state.

The descriptor management has been left the same, trimmed down to only what we need. Descriptors are tracked automatically using the timeline semaphore and are re-used when available. Writes to descriptors are also batched and only performed on scheduler queue submissions. The old texture cache, while more optimal, has been replaced with a commonized one that works for both backends. Further improvements to texture cache can occur in that file that will benefit both backends. The texture decoding code has been pulled out of the OpenGL classes and commonized to avoid duplicating it in the Vulkan side.

The ubershader has been tweaked to use a UBO. The backend does not pass the raw PICA register state like OpenGL, but will perform necessary state decoding on the CPU and send ready uniforms to the GPU. This reduces the amount of data we send each draw by half and makes the shader do less work decoding the data, avoiding potential driver bugs. Depth management is also different, Vulkan has a [0, 1] NDC range which is a perfect mirror of the PICA [-1, 0], so we only need to flip the depth sign.

Data streaming is done using a simplified stream buffer. The buffer enforces a certain granularity on mappings and tracks each region individually using the timeline semaphore. If the buffer cannot fit the data of a certain frame, the scheduler will flush the current command buffer and wait for its execution before mapping again. This is slow but should not happen often (only noted during scene transitions where many textures are uploaded at once and fill the data streaming buffer)

There are still quite a few visual glitches and bugs. Some screen bugs seem to stem from the different coordinate origin points between OpenGL and Vulkan, but these can be resolved with a later PR. Backend has been tested on RADV and the NVIDIA linux driver under Wayland.

Mario Kart 7 | Ocarine of Time 3D | Captain Toad | Mario 3D Land
-|-|-|-
kart | oot3d | toad | 3dland

### DIFF
--- a/include/panda_qt/main_window.hpp
+++ b/include/panda_qt/main_window.hpp
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <SDL.h>
+
 #include <QApplication>
 #include <QMenuBar>
 #include <QtWidgets>
@@ -13,8 +15,8 @@
 #include "emulator.hpp"
 #include "input_mappings.hpp"
 #include "panda_qt/about_window.hpp"
-#include "panda_qt/config_window.hpp"
 #include "panda_qt/cheats_window.hpp"
+#include "panda_qt/config_window.hpp"
 #include "panda_qt/screen.hpp"
 #include "panda_qt/text_editor.hpp"
 #include "services/hid.hpp"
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
wheremyfoodat, line 1: Can this be removed

include/hash.hpp
wheremyfoodat, line 1: This one should probably be yeeted too

PR #393, opened 2024-01-29
add FPS Counter (may work)

Ishan09811:


wheremyfoodat:
Can't we use @GabrielBRDeveloper's frame counter instead since I'm pretty sure that one can actually be turned off and is actually tested ("may work" does not sound promising, please stop making untested PRs)

Ishan09811:
@wheremyfoodat 
ok

PR #400, opened 2024-02-10
Pandroid: check rom file is valid or not

Ishan09811:


wheremyfoodat:
This check is both wrong and already handled by LoadROM

Ishan09811:
> This check is both wrong and already handled by LoadROM

This both check is working fine

wheremyfoodat:
It's absolutely not working fine. It misses most executable types while include .cia which is an installable file and not an executable one ("ROM')... and also not supported yet

To perform a check that already happens in the core

Ishan09811:
> It's absolutely not working fine. It misses most executable types while include .cia which is an installable file and not an executable one ("ROM')... and also not supported yet
> 
> To perform a check that already happens in the core

I don't know much about core but i tried the normal build and i already see the core check is not working and all files is selectable so this check is working fine if you want to remove the he cia then you can.

Ishan09811:
@wheremyfoodat 
You can use this method because this method is working fine (tested).

twvd:
As someone less familiar with 3DS executables, I see that it would refuse to load valid ELF executables (which don't have a fixed file extension, by the way).

Ishan09811:
@wheremyfoodat 
you can just remove the cia from this pr and merge it

wheremyfoodat:
This has the following problems
- Doesn't properly detect all executable types
- Doesn't error on malformed ROMs and the like
- Leads to code duplication, so any time a new ROM type is implemented in the corethen someone will need to change this random line in a random file in the Android code
- Is duplicate code in itself, as this check and a LOT more are in the core's LoadROM function which returns whether ROM loading succeeded

This is why #401 uses LoadROM instead of this

steveplak:
As someone way less familiar with 3DS executables this isssue is irrelevant since
1) the program works already so no need to change it (so you are both wrong)
2) your opinion doesnt matter if your iq is under 3
3) Didnt ask..
Thank you for you time
Kiran
Dis you?

Ishan09811:
@steveplak 
wait a minute who are you ?

PR #408, opened 2024-02-15
Pandroid: Storage Access Framework (SAF)

GabrielBRDeveloper:
# What is this:

In more recents android version, android dont like fs/io operations for user storage, Google recomends use [SAF](https://developer.android.com/guide/topics/providers/document-provider) for manage all files, in moment SAF is many used in java code of this project, but not for load rom files, this pr migrate rom load to SAF implementation.

(thanks for yuzu developers!!)

### DIFF
--- a/src/jni_driver.cpp
+++ b/src/jni_driver.cpp
@@ -15,6 +15,9 @@ RendererGL* renderer = nullptr;
 bool romLoaded = false;
 JavaVM* jvm = nullptr;
 
+jclass alberClass;
+jmethodID alberClassOpenDocument;
+
 #define AlberFunction(type, name) JNIEXPORT type JNICALL Java_com_panda3ds_pandroid_AlberDriver_##name
 
 void throwException(JNIEnv* env, const char* message) {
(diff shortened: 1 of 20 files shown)


src/jni_driver.cpp
```diff
@@ -33,6 +34,23 @@ JNIEnv* jniEnv() {
 	return env;
 }
 
+int AndroidUtils::openDocument(const char* path, const char* perms) {
+    auto env = jniEnv();
+
+    auto clazz = env->FindClass("com/panda3ds/pandroid/AlberDriver");
+    auto method = env->GetStaticMethodID(clazz, "openDocument", "(Ljava/lang/String;Ljava/lang/String;)I");
```
t895, line 41: Maybe not in this PR, but I recommend that you collect method ids once during startup so you're not finding classes each time you want to open a document. It can become pretty costly.

src/pandroid/app/src/main/AndroidManifest.xml
```diff
@@ -2,10 +2,6 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools">
 
-    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
```
wheremyfoodat, line 5: Removing this will break ELF/AXF loading on older devices (while still keeping it broken on newer devices)
Refer to https://discord.com/channels/1118695732958994532/1118702196851216404/1202987391611306044

GabrielBRDeveloper: Fixed!

src/jni_driver.cpp
GabrielBRDeveloper: Implemented, thanks!

PR #410, opened 2024-02-17
WIP: DSP support

wheremyfoodat:
This PR is the first step for full audio support and for fixing games that are still broken due to needing feedback from the DSP (eg Pokemon XY, Yoshi's Island, Prof Layton, Sonic LW, etc)

The PR revamps the way the DSP interface works by adding a nice interface allowing for easily swappable DSP cores. Currently 2 cores are provided: A "null" core (which is the same as current upstream, fast but no audio) and an LLE core using Teakra. HLE DSP core is going to come at a later date (and potentially more DSP cores for trying out different emulation methods).

Credits go to @PSI-Rockin for making the initial version of this DSP interface and starting Teakra integration, and Citra since its Teakra integration code helped a lot

The LLE core is already booting games, but I'll merge the PR after I fix some broken games and make a proper way of swapping between cores that doesn't have to happen at compile time.
https://media.discordapp.net/attachments/845022869007368252/1208539104996761671/image.png?ex=65e3a6c0&is=65d131c0&hm=9f69b50ff9cd8108b7e0edba5f86f1ba8b98c419382a006fdfc2b7fae1d54ae2&=&format=webp&quality=lossless&width=550&height=700.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -55,3 +55,6 @@
 [submodule "third_party/libuv"]
 	path = third_party/libuv
 	url = https://github.com/libuv/libuv
+[submodule "third_party/teakra"]
+	path = third_party/teakra
+	url = https://github.com/wwylele/teakra
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,6 +89,7 @@ include_directories(third_party/toml11)
 include_directories(third_party/glm)
 
 add_subdirectory(third_party/cmrc)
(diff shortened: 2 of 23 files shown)


wheremyfoodat:
Have some music dumped from Yoshi's Island in the LLE backend
[yoshi_long.zip](https://github.com/wheremyfoodat/Panda3DS/files/14335113/yoshi_long.zip)

PR #411, opened 2024-02-18
Improve ``LICENSE``

Ishan09811:


wheremyfoodat:
...You're not meant to change established license text without a good reason, and "running the text under a text formatter" is not a good reason in any form or shape (especially when it's not a proper markdown file). There's a reason GNU gives you the exact license file to use: <https://www.gnu.org/licenses/gpl-3.0.txt>

eepykami:
this is one of the commits of all time

PR #447, opened 2024-03-07
Partially implement ExtSaveData

yretenai:
Refactors SDMC ExtData layout to reflect [VSXE behavior](https://www.3dbrew.org/wiki/Extdata#Virtual_File_System_Structure) and properly stores FormatInfo. 

Created DeleteDirectory and DeleteDirectoryRecursively stubs for future use.

Things that are still a ?:

- "boss" (spotpass) files
- NAND ExtSave (will this just overwrite other title save files?)
- icon

### DIFF
--- a/include/services/fs.hpp
+++ b/include/services/fs.hpp
@@ -34,12 +34,12 @@ class FSService {
 	UserSaveDataArchive userSaveData2;
 
 	std::unordered_map<u64, ExtSaveDataArchive> extSaveData_sdmc;
-	std::unordered_map<u64, ExtSaveDataArchive> sharedExtSaveData_nand;
+	std::unordered_map<u64, ExtSaveDataArchive> nandExtSaveData_nand;
 	SystemSaveDataArchive systemSaveData;
 
 	ArchiveBase* getArchiveFromID(u32 id, const FSPath& archivePath);
-	ExtSaveDataArchive* getExtArchiveFromID(u64 saveId);
-	ExtSaveDataArchive* getSharedExtArchiveFromID(u64 saveId);
+	ExtSaveDataArchive* getExtArchiveFromID(u64 saveId, bool isShared);
+	ExtSaveDataArchive* getNANDExtArchiveFromID(u64 saveId, bool isShared);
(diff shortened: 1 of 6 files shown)


src/core/services/fs.cpp
```diff
@@ -431,11 +432,23 @@ void FSService::deleteDirectory(u32 messagePointer) {
 	const u32 filePathPointer = mem.read32(messagePointer + 28);
 	log("FS::DeleteDirectory\n");
 
-	Helpers::warn("Stubbed FS::DeleteDirectory call!");
+	Helpers::warn("Stubbed FS::DeleteDirectory call!"); // note: should we ensure the system isn't about to delete things outside of the VFS?
```
wheremyfoodat, line 435: (This is done by isPathSafe in the archive implementations)

yretenai: Oh

wheremyfoodat:
Haven't looked into it further yet but I notice a regression in Kirby Planet Robobot which went from working to panicking before booting.

wheremyfoodat:
> Haven't looked into it further yet but I notice a regression in Kirby Planet Robobot which went from working to panicking before booting.

Still haven't tried newer builds, but the commit here had also broken other games like Lego Avengers.

wheremyfoodat:
Can confirm the breakage is still present in aforementioned games + Legend of Zelda: A Link Between Worlds, Rhythm Paradise Megamix and probably more

include/services/fs.hpp
```diff
@@ -33,11 +33,13 @@ class FSService {
 	UserSaveDataArchive userSaveData1;
 	UserSaveDataArchive userSaveData2;
 
-	ExtSaveDataArchive extSaveData_sdmc;
-	ExtSaveDataArchive sharedExtSaveData_nand;
+	std::unordered_map<u64, ExtSaveDataArchive> extSaveData_sdmc;
+	std::unordered_map<u64, ExtSaveDataArchive> nandExtSaveData_nand;
 	SystemSaveDataArchive systemSaveData;
```
wheremyfoodat, line 38: That's a lotta nand

wheremyfoodat: These are also not correctly reset

PR #450, opened 2024-03-08
keep screen on in ``gameactivity``

Ishan09811:


GabrielBRDeveloper:
that looks right @wheremyfoodat

PR #452, opened 2024-03-09
Stub ALL getNsDataIdListX functions

SpikeHD:
Simply stubs all getNsDataIdList functions, instead of just the first two ([3](https://www.3dbrew.org/wiki/BOSS:GetNsDataIdList2) and [4](https://www.3dbrew.org/wiki/BOSS:GetNsDataIdList3)). Gets "Devil Survivor 2" (Japan) to the first screen before having it freak out due to unimplemented Y2R functions 😛, probably helps other games


wheremyfoodat:
Thank you
miss-kobayashi-dragon-maid-tooru

PR #453, opened 2024-03-09
Implement texture filters in GL renderer

wheremyfoodat:


src/core/renderer_gl/renderer_gl.cpp
```diff
@@ -616,7 +616,15 @@ OpenGL::Texture RendererGL::getTexture(Texture& tex) {
 	if (buffer.has_value()) {
 		return buffer.value().get().texture;
 	} else {
-		const auto textureData = std::span{gpu.getPointerPhys<u8>(tex.location), tex.sizeInBytes()};  // Get pointer to the texture data in 3DS memory
+		// Get pointer to the texture data in 3DS memory
+		const auto textureData = std::span{gpu.getPointerPhys<u8>(tex.location), tex.sizeInBytes()};
+		// Override texture filtering if necessary by adjust the texunit config before uploading the texture
+		if (filterSetting == TextureFilter::ForceNearest) {
```
wheremyfoodat, line 622: Mm this should be moved elsewhere before merging, probably in SurfaceCache::Add, cause changing the texture config is very not ideal.

PR #461, opened 2024-03-13
Remove duplicate checks in CMakeLists

OFFTKP:
Because we do add_library near the end, there's a lot of duplicate checks that go like this

```
if (X)
    ** do find_package and stuff like that**
endif()

add_library()

if (X)
    ** link, compile definitions, other stuff **
endif()
```

By moving add_library at the top a lot of these can be merged into one.

Also made it so we don't compile SDL2 on Android when we don't use it there

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,6 @@ if(NOT CMAKE_BUILD_TYPE)
 endif()
 
 project(Alber)
-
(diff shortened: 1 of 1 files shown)


wheremyfoodat:
First thing I notice is that this will totally break Visual Studio filters due to changing how ALL_SOURCES is handled

wheremyfoodat:
Filters seem to work, RIP anyone using CMake 3.10, merging

PR #462, opened 2024-03-14
More CMake cleanups

OFFTKP:
Merge most ENABLE_QT_GUIs, and squash some ifs in cmake:

```
if (STUFF)
    target_compile_definitions(... PANDA3DS_STUFF=1)
endif()
```
to
```
target_compile_definitions(... $<$<BOOL:${STUFF}>:PANDA3DS_STUFF=1>)
```

Generator expressions are described here: https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html

But the gist of it is this:
```
$<condition, true_string>
```
And this string above evaluates to `true_string` if condition == 1, and to nothing if condition == 0

Then, `$<BOOL:string>` does the following
image
so that it can be used in generator expressions, and `$<OR,condition,condition,...>` just does an OR.


OFFTKP:
Removed the generator expression stuff

PR #463, opened 2024-03-14
arm64 shader JIT: Implement PICA200 compliant `MUL`

Wunkolo:
Adds `emitSafeMUL` to implement a PICA200 compliant multiplication that handles the special `0 * inf = 0` case.

Passes unit tests


wheremyfoodat (approved):

wheremyfoodat:
pikachu

PR #466, opened 2024-03-16
Rename Emulator::run to FrontendSDL::run

OFFTKP:


src/panda_sdl/frontend_sdl.cpp
```diff
@@ -67,19 +67,27 @@ FrontendSDL::FrontendSDL() {
 }
 
 bool FrontendSDL::loadROM(const std::filesystem::path& path) { return emu.loadROM(path); }
-void FrontendSDL::run() { emu.run(this); }
 
-void Emulator::run(void* frontend) {
-	FrontendSDL* frontendSDL = reinterpret_cast<FrontendSDL*>(frontend);
-	programRunning = true;
+void FrontendSDL::run() {
+	bool programRunning = true;
```
wheremyfoodat, line 72: Throw the declarations in frontend_sdl.hpp cause this will only really work for as long as this is all 1 big function

OFFTKP: is what I did what you mean

wheremyfoodat: Ye (holdingRightClick is missing from the .cpp file but that tehnically doesn't affect anything)

PR #471, opened 2024-03-22
Implement shader-interpreter relative `MOVA` addressing

Wunkolo:
pando

### DIFF
--- a/src/core/PICA/shader_interpreter.cpp
+++ b/src/core/PICA/shader_interpreter.cpp
@@ -119,9 +119,17 @@ u8 PICAShader::getIndexedSource(u32 source, u32 index) {
 		return source;
 
 	switch (index) {
-		case 0: [[likely]] return u8(source);  // No offset applied
-		case 1: return u8(source + addrRegister[0]);
-		case 2: return u8(source + addrRegister[1]);
+		// No offset applied
+		case 0: [[likely]] return u8(source);
+		// Address register
+		case 1:
+		case 2: {
+			const s32 offset = addrRegister[index - 1];
+			if (offset < -128 || offset > 127) [[unlikely]] {
+				return u8(source);
+			}
+			return u8(source + offset);
+		}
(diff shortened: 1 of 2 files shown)


src/core/PICA/shader_interpreter.cpp
```diff
@@ -303,7 +312,6 @@ void PICAShader::mova(u32 instruction) {
 	const u32 src = getBits<12, 7>(instruction);
 	const u32 idx = getBits<19, 2>(instruction);
 
-	if (idx) Helpers::panic("[PICA] MOVA: idx != 0");
```
wheremyfoodat, line 306: Shouldn't this be using getIndexedSource before getSourceSwizzled?

Wunkolo: Oop you're right. I though loading a value into the address register didn't use the previous address register.

src/core/PICA/shader_interpreter.cpp
```diff
@@ -130,15 +138,16 @@ u8 PICAShader::getIndexedSource(u32 source, u32 index) {
...
-	else if (source < 0x20)
+	} else if (source < 0x20) {
 		return tempRegisters[source - 0x10];
-	else if (source <= 0x7f)
-		return floatUniforms[source - 0x20];
-	else {
-		Helpers::warn("[PICA] Unimplemented source value: %X\n", source);
-		return vec4f({f24::zero(), f24::zero(), f24::zero(), f24::zero()});
+	} else {
+		const usize floatIndex = (source - 0x20) & 0x7f;
```
wheremyfoodat, line 146: Is the wrapping behavior that the & 0x7f incurs confirmed somewhere 🤔

Wunkolo: Confirmed here: https://www.3dbrew.org/wiki/GPU/Shader_Instruction_Set#Relative_addressing

PR #472, opened 2024-03-23
Use `std::bitset` for shader bit-iterations

Wunkolo:
Uses `std::bitset` for bool-uniforms and component-mask iteration: https://en.cppreference.com/w/cpp/utility/bitset

### DIFF
--- a/include/PICA/shader.hpp
+++ b/include/PICA/shader.hpp
@@ -1,6 +1,6 @@
 #pragma once
-#include <algorithm>
 #include <array>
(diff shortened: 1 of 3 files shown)


wheremyfoodat:
I'm not sure I feel particularly great about using bitset for the boolean uniforms, especially when they are directly accessed in JIT code and the standard doesn't guarantee a specific size or layout, nor do I really think it reduces the mental overhead for the component mask stuff...

PR #490, opened 2024-04-08
Allow compilation with opengl=off

Coiland:


wheremyfoodat:
This will compile but also introduces behavior which is unwanted (unsupported renderers failing silently without creating a window, no proper fallbacks or warnings in said case, and so on). Sadly not mergeable, but OpenGL-less builds aren't particularly high priority either at least.

PR #492, opened 2024-04-10
make one workflow for multiple platforms build 

Ishan09811:


wheremyfoodat:
(Having them separate is intentional, putting them all in 1 file becomes too overwhelming to work with very soon)

PR #493, opened 2024-04-13
Update window title on QT and SDL2 frontend to 'Panda3DS'

BlurrySquire:
I noticed that the window title is 'Alber' but the project name is 'Panda3DS'. This change is a bit of a nitpick but I believe it is important to maintain consistency in a project especially with naming conventions.


OFFTKP:
Illegal.

BlurrySquire:
Any specific reasons behind this?

wheremyfoodat:
The name Alber is used everywhere in the core in honor & remembrance of our most valuable companion, so we can't quite change it

BlurrySquire:
Ok that's understandable. Thanks for the insight.

PR #497, opened 2024-04-22
[CRO] Fix faulty data relocations

PSI-Rockin:
The old code based the data pointer used for data relocations off the buffer passed by the application. This is incorrect - the old data segment offset needs to be based upon the start of the CRO file. As a result, data relocations were written to the wrong addresses, potentially causing a crash.

Fixes the NULL pointer crash in ORAS. It likely fixes other crashes in games that use CROs, but this PR needs more testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,4 @@ fb.bat
 *.smdh
 
 config.toml
+CMakeSettings.json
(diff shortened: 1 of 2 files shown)


eepykami:
pokemon

wheremyfoodat:
pokemon

PSI-Rockin:
pokemon

Ishan09811:
pokemon

patataofcourse:
forkemon

PR #499, opened 2024-04-26
DSP HLE: Get ADPCM audio decoding working

wheremyfoodat:


wheremyfoodat:
Shoutout to my Git client for murdering this commit history

PR #500, opened 2024-04-28
Integrate Capstone disassembler

wheremyfoodat:
Adds a class for disassembling code of any supported architecture with Capstone and hooks it to the scripting API for debugging uses.
Same class can later be used for implementing a proper disassembler widget in the UI, viewing recompiler output in real time, etc


wheremyfoodat:
Threw in Teakra's disassembler for the DSP too while I'm at it, will be useful to have a disassembly REPL for putting in some work in @SachinV's Ghidra Teak plugin

image
image

PR #503, opened 2024-04-30
HLE DSP: Stub AAC

wheremyfoodat:


src/core/audio/hle_core.cpp
```diff
@@ -488,6 +502,37 @@ namespace Audio {
 		return decodedSamples;
 	}
 
+	void HLE_DSP::handleAACRequest(const AAC::Message& request) {
+		AAC::Message response = {};
+
+		switch (request.command) {
+			case AAC::Command::EncodeDecode:
+				// Dummy response to stop games from hanging
```
wheremyfoodat, line 510: This will lead to a broken/wrong response header (command, mode, etc will be yeeted), must be fixed before merging

PR #504, opened 2024-05-01
Serialization/deserialization of mappings

OFFTKP:


include/input_mappings.hpp
```diff
@@ -15,8 +18,79 @@ struct InputMappings {
 	}
 
 	void setMapping(Scancode scancode, u32 key) { container[scancode] = key; }
+
+	void serialize(const std::filesystem::path& path) const {
```
wheremyfoodat, line 22: Needs a frontend argument string which will dictate which frontend to save mappings for, trying to interpret QtKey values as SDL scancodes or vice versa will likely explode

include/input_mappings.hpp
```diff
@@ -15,8 +18,79 @@ struct InputMappings {
...
+			}
+
+			data["Mappings"][HID::Keys::keyToName(key)].push_back(scancodeToName(scancode));
+		}
+
+		std::ofstream file(path, std::ios::out);
+		file << data;
+	}
+
+	void deserialize(const std::filesystem::path& path) {
```
wheremyfoodat, line 57: Ditto

include/services/hid.hpp
```diff
@@ -32,6 +33,83 @@ namespace HID::Keys {
...
+			case HID::Keys::CirclePadUp:
+				return "CirclePad Up";
+			case HID::Keys::CirclePadDown:
+				return "CirclePad Down";
+			default:
+				return "Unknown key";
+		}
+	}
+
+	constexpr u32 nameToKey(const std::string& name) {
```
wheremyfoodat, line 76: Ditto
Also marking it constexpr might not be ideal since compilers have had problems with std::string in constexpr contexts until recently

wheremyfoodat: Also this isn't case-insensitive, should probably make it case-insensitive like all other similar configurations

include/input_mappings.hpp
```diff
@@ -15,8 +18,79 @@ struct InputMappings {
...
+
+		try {
+			data = toml::parse<toml::preserve_comments, std::map>(path);
+		} catch (const std::exception& ex) {
+			Helpers::warn("Exception trying to parse config file. Exception: %s\n", ex.what());
+			
+			return;
+		}
+
+		name = toml::find_or<std::string>(data, "Metadata.Name", "Unnamed Mappings");
```
wheremyfoodat, line 77: Is the Metadata.Name syntax legal? Can't find it in the docs.

OFFTKP: It is (dotted keys) but it doesn't work like I think it did

PR #512, opened 2024-05-11
Index with iterator value rather than getSingleIndex() in CAMService::startCapture

NerduMiner:
The port may have a value of 3 in this function, which will cause a panic when using getSingleIndex(). getPortIndices() handles this case for us already, so the iterator value is safe to use.

Fixes a crash in The Denpa Men: They Came By Wave (USA).


wheremyfoodat:
Lookin good, thanks!

PR #518, opened 2024-06-04
Fix path sanitization

OFFTKP:


wheremyfoodat:
This should probably just explicitly check for `./` instead. This implementation is broken in a different way (paths of a specific archive can access, eg the SD archive, will be able to access anything inside the emulator folder, even outside the archive)

OFFTKP:
> This should probably just explicitly check for `./` instead and not increment the nesting level in that case. This implementation is broken in a different way (paths of a specific archive can access, eg the SD archive, will be able to access anything inside the emulator folder, even outside the archive)

Yes, although to be fair that was already possible with your implementation and checking for ./ doesn't really fix that. Also checking for ./ isn't enough, what about paths like `///////../../../../../`? The multiple forward slashes are also a potential problem so you'd have to check for that too. Why not use a ready made, safe, standard solution?

I think perhaps the change you want me to make is to make sure `isSafePath` checks the path is inside the current archive instead, which means taking another parameter for the archive directory we want to check for at any given time.

OFFTKP:
> Yes, although to be fair that was already possible with your implementation

For example `/../file.txt` is a safe path with the current implementation, as you add 1 for the forward slash, but that forward slash is just the base of the current archive path

PR #531, opened 2024-07-09
Improve lighting

OFFTKP:
# Uncomment lighting on Android
- Made lightingLUT's type GL_TEXTURE_2D and u_tex_lighting_lut to sampler2D
- Made lightingLUT's internalFormat GL_R32F, and converting the LUT data to floats before uploading.
  Internal format and format types need to match according to the spec:
image
My guess is the conversion to float was happening implicitly in desktop GL at some point, because `lutLookup` returns a float. The type `GL_R16` was probably interpreted as `GL_R16F` ?
The alternative to this could be using GL_R16UI, but then we'd need to convert to float on the GPU 
- Switched to using `texelFetch` instead of `texture` in the shader, since this texture is really treated as an array and we don't need filtering or interpolation
- Updated the gles.patch

# Improve lighting implementation
- Fix light_vector not being properly updated for positional lighting
- Disabled factors were being set to 0 instead of 1 (see Lego Batman's fix)
- GPUREG_LIGHTING_LUTINPUT_ABS was not handled correctly (it's 0 == enabled, abs was on output instead of input, would use wrong bits)
- GPUREG_LIGHTING_LIGHT_PERMUTATION was not handled correctly (it was accessed every 3 bits when it's every 4 bits per enabled light)
- Fix spotlight attenuation, doesn't use LUT 2 but LUT's 8-15
- Fix environment 4 (SP), which uses GPUREG_LIGHTi_SPOTDIR, it's a fixed point number and not a floating point number as it was treated before
- Implement distance attenuation
- Implement geometric factors
- Implement specular clamping
- Clamp results at the end of calcLighting
- Removed unused v_tangent, v_bitangent, but might need them for bump lighting at some point (although they'll be treated differently)
- Removed pointless normalization of v_normal as it already comes in normalized from the vertex shader
- Probably other stuff I forgot

# Known bugs
- Luigi's mansion:
    - ~~Luigi~~ Luigi's lighting is mostly fixed except when he opens the door he looks like he's seen a ghost:
image
    - The birds in the intro are too bright, edit: are they now? unsure
    - ~~Spotlight just decides to stop working sometimes. It continues working when thunder strikes for a short while, then it stops working and the whole room goes brighter~~ Fixed, GPUREG_LIGHTING_LIGHT_PERMUTATION was being accessed incorrectly
    - ~~In the intro when thunder strikes twice the road isn't illuminated~~ Fixed
    
- Cave Story 3D:
    - ~~The evil doctor (and the rest of the scene) in the boot up of the game is wrongly lit. Game is better lit than before overall, was almost completely dark~~ This should be fixed now post the Luigi spotlight fix
    
- Rabbids:
    - ~~Positional lighting seems to work now, however there's some light flickering in the level selector near the castle~~ Seems fixed after the Luigi spotlight fix, can't see any flickering
    
- Captain Toad Treasure Tracker:
    - Entire level is still too bright but better than before
    
- Pokedex 3D:
    - Pokemon not lit correctly, needs bump mapping
    
- Pushmo/Crashmo:
    - Seemingly same light issues as before on the characters

- Lego Batman:
    - There's supposed to be some glare on top of the legos, there isn't:
image

    
# Improvements
- toon_shading.elf and fragment_lighting.elf seem to work fine
- Pokemon Omega Ruby is no longer too bright everywhere, lighting seems fine to me
- Pokemon SMD, same
- Super Mario 3D Land, same
- Cave Story 3D looks fine to me
- Luigi's mansion is better but there's still issues
- Captain Toad Treasure Tracker is better but still too bright
- Kirby Triple Deluxe seems less bright on his head and that looks closer to playthoughs I see online
- Lego Batman is better, should be some specular lighting missing on the lego heads

This patch makes it so that the shader editor shows the LUTs, useful for debugging
[luts.txt](https://github.com/user-attachments/files/16240512/luts.txt)

### DIFF
--- a/src/host_shaders/opengl_fragment_shader.frag
+++ b/src/host_shaders/opengl_fragment_shader.frag
@@ -272,8 +272,8 @@ float lightLutLookup(uint environment_id, uint lut_id, uint light_id, vec3 light
 		}
 		case 4u: {
 			// These are ints so that bitfieldExtract sign extends for us
-			int GPUREG_LIGHTi_SPOTDIR_LOW = int(readPicaReg(0x0146u + 0x10u * light_id));
-			int GPUREG_LIGHTi_SPOTDIR_HIGH = int(readPicaReg(0x0147u + 0x10u * light_id));
+			int GPUREG_LIGHTi_SPOTDIR_LOW = int(readPicaReg(0x0146u + (light_id << 4u)));
+			int GPUREG_LIGHTi_SPOTDIR_HIGH = int(readPicaReg(0x0147u + (light_id << 4u)));
 
 			// These are fixed point 1.1.11 values, so we need to convert them to float
 			float x = float(bitfieldExtract(GPUREG_LIGHTi_SPOTDIR_LOW, 0, 13)) / 2047.0;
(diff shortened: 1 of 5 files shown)


wheremyfoodat:
cc @skylersaleh

wheremyfoodat:
Someone is cooking with this one 👀 (Stolen from Paris' discord messages)

image
image

src/host_shaders/opengl_fragment_shader.frag
```diff
@@ -38,6 +38,21 @@ vec4 tevSources[16];
 vec4 tevNextPreviousBuffer;
 bool tevUnimplementedSourceFlag = false;
 
+// Holds the enabled state of the lighting samples for various PICA configurations
+// As explained in https://www.3dbrew.org/wiki/GPU/Internal_Registers#GPUREG_LIGHTING_CONFIG0
+const bool samplerEnabled[9 * 7] = bool[9 * 7](
```
wheremyfoodat, line 43: Might be better to pack this bools into bitfields

src/host_shaders/opengl_fragment_shader.frag
```diff
@@ -178,157 +199,277 @@ float decodeFP(uint hex, uint E, uint M) {
...
+	}
+
+	uint GPUREG_LIGHTING_AMBIENT = readPicaReg(0x01C0u);
+	vec4 global_ambient = vec4(regToColor(GPUREG_LIGHTING_AMBIENT), 1.0);
+	primary_color = clamp(global_ambient + diffuse_sum, vec4(0.0), vec4(1.0));
+	secondary_color = clamp(specular_sum, vec4(0.0), vec4(1.0));
 
 	if (error_unimpl) {
-		// secondary_color = primary_color = vec4(1.0, 0., 1.0, 1.0);
+		secondary_color = primary_color = unimpl_color;
```
wheremyfoodat, line 472: Probably want to comment out this line too because people will be blinded by magenta everywhere instead of slightly messed up lights

OFFTKP:
Adding a +1 to num lights makes the flashlight light come back
image

image

wheremyfoodat:
Closed this accidentally deleting the other branch

OFFTKP:
Needs vertex shader to be changed in specialized shader PR

src/host_shaders/opengl_fragment_shader.frag
OFFTKP: Donezo

src/host_shaders/opengl_fragment_shader.frag
OFFTKP: Done

OFFTKP:
TODO: bitfieldExtractCompat needs to sign extend for spotdir, or do something else

PR #532, opened 2024-07-11
Fix AES-CTR decryption for non-NCCHKey0 games

wheremyfoodat:
Fixes several encrypted dumps for later games not working, tested with Glory of Generals and Ninja Usagimaru.
Likely fixes other games reported to be broken with encrypted dumps, such as Luigi's Mansion (cc @Adamillo If you could test it pretty please)

### DIFF
--- a/include/loader/ncch.hpp
+++ b/include/loader/ncch.hpp
@@ -60,6 +60,8 @@ struct NCCH {
 	CodeSetInfo text, data, rodata;
 	FSInfo partitionInfo;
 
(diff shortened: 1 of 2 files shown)


Adamillo:
Sure thing! I can test it tomorrow!

wheremyfoodat:
Implementing NCCH version 1 is also a pending issue since it affects the AES IV, though I have still not found a game that uses that. Might score with Luigi's Mansion though if it still doesn't work after this.
image

PR #536, opened 2024-07-14
x64 shader rec: Add support for PICA non-IEEE multiplication

wheremyfoodat:
Adds optional support for safe mul in the x64 shader recompiler for dp3/dp4/dph/mul/mad, passes Wunk's mul testing.


wheremyfoodat:
Fixes OoT bottom screen icons when enabled
image

PR #537, opened 2024-07-16
Add accurate shader multiplication option

wheremyfoodat:
Exposes accurate shader multiplication as an optional setting (off by default) that users can toggle & is applied when the emulator is reset. Useful for specific games like Ocarina of Time 3D which might require the shader units' weird non-IEEE multiplication being emulated properly to render correctly.

Supported in all frontends including Libretro.


wheremyfoodat:
(In the future, there should ideally be overrides using title IDs for deciding whether potentially performance-affecting options such as this should be enabled or not)

PR #541, opened 2024-07-19
Remove dependency of PICA regs in fragment config

OFFTKP:
Should be helpful with needing less copies for when we move to hybrid shaders

### DIFF
--- a/include/PICA/pica_frag_config.hpp
+++ b/include/PICA/pica_frag_config.hpp
@@ -73,18 +73,7 @@ namespace PICA {
 			BitField<22, 2, u32> shadowSelector;
 		};
 
-		LightingLUTConfig d0{};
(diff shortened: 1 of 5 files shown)


include/PICA/pica_frag_config.hpp
```diff
@@ -81,6 +81,11 @@ namespace PICA {
 		LightingLUTConfig rg{};
 		LightingLUTConfig rb{};
 
+		u32 config1;
+		u32 lutAbs;
```
wheremyfoodat, line 85: These are duplicated, the LutConfig members should be changed to be an array and then we can access these attributes from there without duplicating them

wheremyfoodat (approved):
Lmfao squash the PR but don't remove the commit names from the squash comment

PR #543, opened 2024-07-20
Add support for asynchronous shader compilation

OFFTKP:
This allows for using the ubershader while specialized shaders are compiling asynchronously, which should reduce compilation stutter.

See docs/3ds/asynchronouse_shader_compilation.md for more info

WIP

### DIFF
--- a/include/renderer.hpp
+++ b/include/renderer.hpp
@@ -21,6 +21,20 @@ enum class RendererType : s8 {
 };
 
 struct EmulatorConfig;
+enum class ShaderMode {
+	Specialized = 1,
+	Ubershader = 2,
+	Hybrid = 3,
+};
+
+// For now, use specialized shaders by default on MacOS as M1 drivers are buggy when using the ubershader, and on Android since mobile GPUs are
+// horrible. On other platforms we default to ubershader + shadergen fallback for lights
+#if defined(__ANDROID__) || defined(__APPLE__)
+	static constexpr ShaderMode defaultShaderMode = ShaderMode::Specialized;
+#else
+	static constexpr ShaderMode defaultShaderMode = ShaderMode::Ubershader;
+#endif
+
 class GPU;
 struct SDL_Window;
 
@@ -56,6 +70,8 @@ class Renderer {
 	static constexpr u32 vertexBufferSize = 0x10000;
 	static std::optional<RendererType> typeFromString(std::string inString);
 	static const char* typeToString(RendererType rendererType);
+	static std::optional<ShaderMode> shaderModeFromString(std::string inString);
+	static const char* shaderModeToString(ShaderMode shaderMode);
 
 	virtual void reset() = 0;
 	virtual void display() = 0;                                                              // Display the 3DS screen contents to the window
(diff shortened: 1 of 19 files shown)


src/core/renderer_gl/async_compiler.cpp
```diff
@@ -0,0 +1,115 @@
...
+    Stop();
+}
+
+bool AsyncCompilerState::PushFragmentConfig(const PICA::FragmentConfig& config)
+{
+    PICA::FragmentConfig* newConfig = new PICA::FragmentConfig(config);
+    bool pushedSuccessfully = configQueue.Push(newConfig);
+
+    if (!pushedSuccessfully) {
+        Helpers::panic("Hlep we managed to fill the shader queue");
```
wheremyfoodat, line 25: Hlep

include/renderer_gl/async_compiler.hpp
```diff
@@ -0,0 +1,74 @@
...
+        : fsConfig(fsConfig)
+    {}
+
+    PICA::FragmentConfig fsConfig;
+	std::vector<u8> binary;
+	u32 binaryFormat;
+};
+
+namespace PICA
+{
```
wheremyfoodat, line 33: namespace PICA::ShaderGen

include/renderer_gl/async_compiler.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include "helpers.hpp"
+#include "opengl.hpp"
+#include "PICA/pica_frag_config.hpp"
+#include "spsc/queue.hpp"
+
+// We make the assumption that not more than 256 shaders will be queued at once.
+// This may seem like a hack, but it's a reasonable assumption and it allows us to not have
+// to implement a waiting mechanism (wait until the queue is not full)
+// In the unlikely event that either queue is full, a panic will be triggered.
+constexpr size_t maxInFlightShaderCount = 256;
```
wheremyfoodat, line 18: Should be in the async compiler class instead of global

src/core/renderer_gl/async_compiler.cpp
OFFTKP: It's a reference to kernel.hpp:215 !!!

src/core/renderer_gl/async_compiler.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    if (hasItem) {
+        program = newProgram;
+    }
+
+    return hasItem;
+}
+
+void AsyncCompilerState::createGLContext() {
+    // TODO: do me
```
wheremyfoodat, line 44: 😳

src/core/renderer_gl/async_compiler.cpp
wheremyfoodat: Ik it's perfect

include/renderer.hpp
```diff
@@ -21,6 +21,20 @@ enum class RendererType : s8 {
 };
 
 struct EmulatorConfig;
+enum class ShaderMode : s8 {
```
wheremyfoodat, line 24: (It's probably better to not have a type specification here and let the compiler pick, not like this is really getting used as a numerical value)

OFFTKP:
Closing this, going to open a new PR in the future with shared contexts, not overriding it in case the code here is needed in the future

PR #544, opened 2024-07-20
Libretro: Add support for touch input

jonian:
This add support for touchscreen input. Do you think that they should be toggle-able with settings like in the citra core?
```cpp
{"panda3ds_mouse_touchscreen", "Simulate touchscreen with mouse; enabled|disabled"},
{"panda3ds_touch_touchscreen", "Simulate touchscreen with touch; enabled|disabled"},
```
I can also add support for joystick emulation but that needs to render a pointer and I wasn't able to do it by following an implementation like that in citra core [here](https://github.com/libretro/citra/blob/2d67658e85de4767c0eefeb2829d710401c5c802/src/citra_libretro/input/mouse_tracker.cpp#L184).

### DIFF
--- a/src/libretro_core.cpp
+++ b/src/libretro_core.cpp
@@ -15,6 +15,8 @@ static retro_input_state_t inputStateCallback;
 static retro_hw_render_callback hw_render;
 static std::filesystem::path savePath;
 
+static bool screenTouched;
+
 std::unique_ptr<Emulator> emulator;
 RendererGL* renderer;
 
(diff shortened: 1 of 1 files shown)


wheremyfoodat:
> Do you think that they should be toggle-able with settings like in the citra core?
I've got a touchscreen laptop and I've always found it comfortable to be able to press things with both my screen and my mouse at the same time, so I'd say nah it's fine. Don't think anyone minds unless they're on a phone with a mouse at which point they're probably suffering anyways.

PR #545, opened 2024-07-21
Fix shadowed variable in ubershader

OFFTKP:
This PR removes Luigi's Latex suit in Luigi's Mansion
Before:
image
After:
image


wheremyfoodat:
6aee3cee-7ef5-495f-8a79-e56f5879ee53_text

PR #560, opened 2024-07-25
Qt: Stop emuThread on closeEvent

jonian:
This moves emuThread stopping to main window closeEvent. Without this the app keeps running even if the window is closed. Also added `SDL_Quit`, but I'm not sure if it is necessary.


wheremyfoodat:
Removed SDL_Quit as SDL is vague about whether you can SDL_Quit from another thread than the initialization one, shouldn't matter though. I didn't put it in the emulator thread cause that code may or may not change depending on another branch. Otherwise lgtm.

jonian:
Thank you for taking the time to explain things!

PR #561, opened 2024-07-25
Add SDL resizing & optimize GL shadergen

wheremyfoodat:


wheremyfoodat:
Moving fog colour to a uniform instead of FSConfig is also necessary for Majora's Mask 3D, as this scene in particular constantly shifts the fog colour and will otherwise generate thousands of shaders
image

PR #576, opened 2024-08-14
Implement controller-based gyroscope on both Qt and SDL

wheremyfoodat:
Allows players to control the 3DS gyroscope by moving around their controller, provided their controller has a gyroscope sensor. Makes games like Kirby's Triple Deluxe, Yoshi's New Island and Pokemon X, all of which require gyroscope to some degree, a lot more playable.

Tested with my Sony DualShock 4 on Windows.


OFFTKP:
image

OFFTKP (approved):

PR #587, opened 2024-08-27
macOS CI fixes

shinra-electric:
This PR updates the macOS build yaml with the following changes:
- Update `checkout` from v2 to v4
- Update `upload-artifact` from v2 to v4
- Removes install step for Python 3.12

Reason for changes: Resolving deprecation warnings
- v2 of `checkout` and `upload-artifact` are deprecated. 
- Python 3.12 is already installed and up to date

Testing: 
- Ensure CI still builds

### DIFF
--- a/.github/workflows/MacOS_Build.yml
+++ b/.github/workflows/MacOS_Build.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: macos-13
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


shinra-electric:
I made the same changes to `Qt_Build.yml` too, but I didn't update the Windows or Linux jobs because this PR is supposed to be for macOS. They should be updated too. 

There is still a deprecation warning for `cache@v2` somewhere that needs to be updated, but I haven't found it yet...

PR #613, opened 2024-10-20
GPU: Add optimized NEON path for analyzing index buffers

wheremyfoodat:
Should improve perf on arm64 devices. I have written most of an x64 version too, but it requires SSE4.1 so I am unsure how to proceed on that.


wheremyfoodat:
Added SSE4.1 path and enabled SSE4.1 support for GCC/Clang builds by default unless disabled via the DISABLE_SSE4 option, sadly can't do the same for MSVC because there's no /arch:SSE4.1 flag, only SSE4.2

Would be ideal to switch to clang-cl but annoyingly, cryptopp refuses to work there

PR #617, opened 2024-10-23
GL renderer: Add fallback for when driver doesn't provide glDrawRangeElementsBaseVertex

wheremyfoodat:


wheremyfoodat:
Should fix crash on Android Studio emulator and mobile devices that support GLES 3.1 and not 3.2

PR #631, opened 2024-11-06
Add rsyn emote-based png and ico files for future use

extherian:
Extra icons for use with the hypothetical icon-swapping feature within the app that was mentioned on the discord. 

These files are based on the rsyn emote from the discord, flipped horizontally and edited in gimp to remove the blobs of colour from the corners. Although they are not visible on the discord emote, these are very noticeable when used as a desktop shortcut. rsyn.png could be used as an image within the app itself in place of rpog.png, which appears in the title bar, as well as for use as a desktop shortcut. windows_alt_icon.ico could be used for a rainbow-coloured version of the existing windows_icon.ico file, though swapping this out would be harder. Perhaps some future Panda3DS installer could swap it in place of the existing ico file during the installation process if desired. 

Resolution and aspect ratio are identical to the existing rpog.png and windows_icon images.


wheremyfoodat:
Awesome

PR #632, opened 2024-11-10
WIP: Finishing DSP mixer

wheremyfoodat:


wheremyfoodat:
(Ideally this will be merged soon enough when dumped audio is 99% correct and as many hot paths as possible have SSE4.1/NEON paths in order to minimize performance penalty. Hooking the DSP to miniaudio can come a little bit later in a separate PR)

PR #648, opened 2024-11-28
Maintain device resolution on window resize for OpenGl

Coiland:


wheremyfoodat:
Away from computer atm, this seems like it should work from a math perspective but I'll take some time to clean it up & hook it to a "maintain aspect ratio" setting in the emulator settings.

PR #669, opened 2024-12-06
Qt: Add Spanish translation

DaniElectra:
Every string is translated except for two of the color themes `Cream` and `Greetings Cat` as I'm unsure where to literally translate it or keep the original name.


wheremyfoodat:
Gracias 🥺

PR #683, opened 2024-12-10
Fix building crypto++ for x64 target on Apple silicon MacOS

twvd:
I think this will let you use the M1 Github runners, try it..


wheremyfoodat:
Won't this break if you're on Apple and CMAKE_OSX_ARCHITECTURES is not defined?

twvd:
True.. Fixed that

wheremyfoodat:
~~What if I'm on Windows but I'm a lunatic that defines CMAKE_OSX_ARCHITECTURES /s~~

wheremyfoodat (approved):

PR #692, opened 2024-12-26
Decouple data save paths from Emulator

twvd:
This change allows the frontend to select the appdata and configuration file locations when instantiating the emulator, allowing for more flexibility in placing the files.

This also fixes configuration not working in MacOS app bundles, as it will now look for config.toml in the user's "Application Support" directory and default there.

### DIFF
--- a/include/emulator.hpp
+++ b/include/emulator.hpp
@@ -28,6 +28,8 @@
 #include "gl/context.h"
 #endif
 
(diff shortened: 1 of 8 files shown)


wheremyfoodat:
This also fixes launching games from the Windows start menu or when you've configured Alber to open 3DS files (on upstream, this uses C:/System32 for the config file, which it doesn't even have write access to)

PR #701, opened 2024-12-29
Fix git versioning

jonian:
Fixes versioning from git when git tags are missing (eg. CI)


wheremyfoodat:
I don't know what these words mean but I trust you with my life

wheremyfoodat (approved):

PR #763, opened 2025-06-29
SDL/Qt: Better resizing & fullscreen support

wheremyfoodat:
TODO:
- ~~Clear the window before drawing the output since we're not doing a fullscreen draw anymore~~
- Make sure there's no multithreading race conditions in the Qt frontend


wheremyfoodat:
Ah I had forgotten I already do clearing, so there goes that
image

PR #770, opened 2025-07-04
Lua services interception upgrade

LiquidFenrir:
That feature was great, good work! Some design decisions werent my favourite though, as we discussed on discord, so I'm PRing this, see if you want it

- Faster and better lookup of the service names from handles should improve the frame drops you noticed sometimes
  - first commit: prevent unnecessary string copies
  - second commit: use a reverse form of serviceMap so the better algorithm from the data structure kicks in
- Pand.addServiceIntercept in the Lua code now takes a third callback argument, so "interceptService" in the Lua code is no longer implicitly the handler for all the intercepts
  - Easier to make separate handlers for separate things, or even anonymous ones (see comments in lua.cpp:addServiceInterceptThunk)
  - This should also make the intercepts faster because of looking up an integer key in the registry instead of a string key in the global scope

### DIFF
--- a/include/lua_manager.hpp
+++ b/include/lua_manager.hpp
@@ -48,7 +48,8 @@ class LuaManager {
 		}
 	}
 
-	bool signalInterceptedService(const std::string& service, u32 function, u32 messagePointer);
+	bool signalInterceptedService(int callback_ref, const std::string& service, u32 function, u32 messagePointer);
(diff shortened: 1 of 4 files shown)


wheremyfoodat:
Thanks, works great! I moved the PR to #770 adding you as co-author because I committed horrible git atrocities and ruined the git commit history while trying to fix a merge conflict.

PR #787, opened 2025-07-23
GPU: Fix VertexShaderOutputMask in non-accelerated draws

wheremyfoodat:
This is a regression that seems to have been introduced when adding hw shaders. Fixes VertexShaderOutputMask emulation in the shader interpreter & CPU JIT backends.

This fixes various games that regressed when not using hw shaders, such as Pokemon Super Mystery Dungeon.


wheremyfoodat:
image

PR #828, opened 2026-07-14
Set CI build to use Windows-2022 instead of latest

wheremyfoodat:
Debugging an issue with latest Windows builds not properly loading games


wheremyfoodat:
Confirmed to fix Windows regressions, urgent patch applied directly to master.

PR #830, opened 2026-07-28
ci: add Linux aarch64 and Android x86_64 builds (+ fix cryptopp on ARMv8)

WizzardSK:
Two targets the libretro buildbot does not currently ship.

**Linux aarch64** — the job needs cryptopp's SIMD sources to be given their `-march` flags on ARMv8, otherwise the build fails on the NEON/crypto intrinsics; that fix is the second commit.

**Android x86_64** — enables the job that was already sketched out (commented) in `.gitlab-ci.yml`, mirroring arm64-v8a with the same NDK pin. It is the ABI used by Android emulators and x86 Chromebooks. 32-bit x86 stays commented out: the core needs a 64-bit address space.

Rebased on current master.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -62,6 +62,18 @@ libretro-build-linux-x64:
     - .libretro-linux-cmake-x86_64
     - .core-defs
 
+# Linux aarch64
+# No image override here: the :backports tag exists only for the amd64 image,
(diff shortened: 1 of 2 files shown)


wheremyfoodat:
Paging @warmenhoven since they know all the libretro stuff

WizzardSK:
Thank you very much.

PR #831, opened 2026-07-29
Switch to Azahar's Dynarmic

RedBlackAka:
Switch to [Azahar's Dynarmic fork](https://github.com/azahar-emu/dynarmic), containing some minor optimizations and centralizing future development. Notably includes a change to not use a random number generator on ARM, resulting in a small performance improvement for those devices

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -79,4 +79,4 @@
 	url = https://github.com/Panda3DS-emu/xbyak
 [submodule "third_party/dynarmic"]
 	path = third_party/dynarmic
-	url = https://github.com/panda3ds-emu/dynarmic.git
+	url = https://github.com/azahar-emu/dynarmic.git


wheremyfoodat:
Thanks. CI seems to pass so I'll be merging it. But since it might need some more testing on certain platforms, what have you tested it on?

wheremyfoodat (approved):

RedBlackAka:
> Thanks. CI seems to pass so I'll be merging it. But since it might need some more testing on certain platforms, what have you tested it on?

Nice, tested on Windows and Android with ACNL and OoT, with no regressions compared to latest master. As noted, our branch has relatively minor changes compared to yours, nothing to break here atm

wheremyfoodat:
Fire

PR #832, opened 2026-07-30
CI: Manually set macOS deployment target to 13.3

RedBlackAka:
The GitHub workflow runners seem to not respect the macOS minimum deployment target set in CMakeLists.txt, so we can manually set it in the workflow files to make it apply. Fixes the Panda3DS .app bundle throwing an error and refusing to launch on older OSes beneath 14.0, tested specifically with 13.7


wheremyfoodat:
Thanks!!

wheremyfoodat (approved):

PR #834, opened 2026-08-06
Android_Build: refactor

Ishan09811:


wheremyfoodat:
The build artifact formatting broke, now it's `pandroid-(arm64)` instead of `pandroid-arm64`

Ishan09811:
> The build artifact formatting broke, now it's `pandroid-(arm64)` instead of `pandroid-arm64`

is it even a problem?😑

wheremyfoodat (approved):

wheremyfoodat:
cc @GabrielBRDeveloper in case he has any disagreements

Ishan09811:
> cc @GabrielBRDeveloper in case he has any disagreements

why he will have any disagreements regarding something he didn't even made 😑

wheremyfoodat:
> > cc @GabrielBRDeveloper in case he has any disagreements
> 
> why he will have any disagreements regarding something he didn't even made 😑

Well that's kind of the way reviews work, we do not tend to disagree with ourselves very often.

Ishan09811:
> > > cc @GabrielBRDeveloper in case he has any disagreements
> > 
> > 
> > why he will have any disagreements regarding something he didn't even made 😑
> 
> Well that's kind of the way reviews work, we do not tend to disagree with ourselves very often.

ok but it is unnecessary because he will agree after all

PR #835, opened 2026-08-10
Fix Qt6.10+

neptuwunium:
fixes `src/panda_qt/screen/screen.cpp:14:10: fatal error: 'qpa/qplatformnativeinterface.h' file not found`

this is due to that header existing on `/usr/include/QtGui/6.11.1/QtGui/qpa/qplatformnativeinterface.h` which is not in the search paths otherwise.


wheremyfoodat (dismissed):

wheremyfoodat:
Thanks :bow:

neptuwunium:
OK Looking into the error, it's because Qt 6.10  introduced a breaking change moving a bunch of stuff to *Private libraries. install-qt-action installs the latest LTS (6.8). Next LTS is 6.12 which releases end of September.

Either defer this patch until 6.12 is released (and present in distros) or rely on Qt 6.10+

neptuwunium:
After sleeping on it I realized we could just do `if(Qt6_VERSION VERSION_GREATER_EQUAL "6.10.0")` which solves the entire conundrum.

